PR #85, opened 2014-07-09
Adds a MySQL Dynamic

jpbarraca:
This service is similar to the provided MySQL service but is able to create tables on the fly as data is published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,13 +615,39 @@ to have those values stored automatically.
 
 ### `mysql_dynamic`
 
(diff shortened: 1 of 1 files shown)


vi1004t:
Hello! I tried to configure the mysql_dynamic with the example:

mosquitto_pub -t my/2 -m '{ "name" : "Jane Jolie", "id" : 90, "number" : 17 }'

But mqttwarn don't work. Don't work because the example duplicates columns ('id' data). Please update the example and include this information, and mention which columns can not be used because are reserved by system (id, _dtepoch, _dtiso, _dthhmm, payload, _dthhmmss).

I hope this information is helpful

Issue #121, opened 2015-04-18
How to pass received mqtt payload to plugin (instapush)


pfichtner:
How can i pass the payload received to the instapush plugin. 

```
[foo/bar]
targets = instapush:webcamalert

[config:instapush]
appid = 'secret'
appsecret = 'secretToo'
targets = {
  'webcamalert' : [ 'webcam', {"device":"hardCodedDevice", "type":"hardcodedType"}]
  }
```

When publishing a message via ```mosquitto_pub -h mqtt -t foo/bar -m '{"device":"webcam 4", "type": "motion"}'```

I'd like to send the payload received by mqtt (```{"device":"Webcam #4", "type": "motion"}```) 
and not the static content ```{"device":"hardCodedDevice", "type":"hardcodedType"}```

I played around with format but couldn't get it to work

Thank you!

jpmens:
What should work is using the _format_ specifier as follows:

```ini
format = {payload}
```

pfichtner:
and how would  ```targets = {
  'webcamalert' : [ 'webcam', {"?????"}]
  }``` of ```[config:instapush]``` look like?

jpmens:
That _looks_ right to me, but I have no experience with instapush.
@dgomes Can you give us a hand here?

pfichtner:
Just for clarification: I do know how to format the payload via ```format``` but I don't know what to write inside the plugin's configuration to access the formatted string (see the question marks in my last post)

pfichtner:
Any news here? Is this a general issue or a plugin related issue? How can a plugin access the string formatted by ```format```? Does each plugin have to implement this by it's own?

jpmens:
I'm losing track a bit ... Plugins get the text formatted by `format` automatically.

However: glancing at the source of the `instapush` plugin, I see it isn't being used at all; instead the author (@dgomes) is using the parameters in the target configuration, which is a bit "unusual" maybe. ;-)

If you need this quickly, you're going to have to copy `instapush.py` and call it something else (and modify it accordingly). It's not hard: the `file.py` module shows you how to get at the _text_.

pfichtner:
Oh! Alright! No need to skelter! Sufficient to know it wasn't my fault and couldn't be fixed by me on configuration level! 
Will wait for @dgomes to fix it! 
Thank you!

jpmens:
[...]

PR #123, opened 2015-04-23
Instapush service update

dgomes:
Attends to #121


jpmens:
Merged, thanks. (Untested!)

PR #124, opened 2015-04-26
new service: MythTV service

stefanroellin:
This service allows for on-screen notification popups on [MythTV](http://www.mythtv.org/) instances.


jpmens:
Merged, thank you!

PR #126, opened 2015-05-03
warn if cleansession is false

AnDann:
I think this little warning could be helpful while setting everything up.


jpmens:
Merged, thank you. Yes, probably a good idea. I'll change the text slightly.

Issue #128, opened 2015-05-10
Remove osxnotify from sample ini


robertvanbregt:
Remove osxnotify from sample ini, since it does not run out of the box. It requires

* a Mac :)
* additional installation

robertvanbregt:
I will create a pull request

PR #129, opened 2015-05-10
Remove osxnotify from sample ini

robertvanbregt:
PR for #128


jpmens:
Merged, thanks. Probably a good idea.

PR #134, opened 2015-05-28
fix OwnTracksBattFilter function

henkjan:
WARNING [mqttwarn] Cannot invoke filter function OwnTracksBattFilter defined in owntracks-battery: local variable 'data' referenced before assignment


jpmens:
Whoops: I'm sorry about that!
Merged, thanks!

PR #142, opened 2015-07-17
Add function for nsca

cuonglv1994:
This commit add function for nsca, first convert payload to dictionary
type, then work on it to get service description and host name for nsca
service. I hope you could somehow use this piece of code.

def nsca_data_dict(payload):
	try:
		converted_payload = "{"+payload+"}"
		converted_payload = ast.literal_eval(converted_payload)
	except:
		logging.debug("payload content: %s is not valid for parsing", payload)
	return converted_payload

in processor() function:
if ( service == "nsca" ):
				payload = nsca_data_dict(job.payload)
				if (payload.has_key('HostName')):
					item = {
						'service'       : service,
						'section'       : section,
						'target'        : target,
						'config'        : get_service_config(service),
						'addrs'         : [payload['HostName'], payload['Name']],
						'topic'         : job.topic,
						'payload'       : payload,
						'data'          : None,
						'title'         : None,
						'image'         : None,
						'message'       : None,
						'priority'      : None
					}
					logging.debug("item content: `%s'" % (item))

### DIFF
--- a/mqttwarn.py
+++ b/mqttwarn.py
@@ -165,16 +165,12 @@ class Config(RawConfigParser):
     def config(self, section):
         ''' Convert a whole section's options (except the options specified
             explicitly below) into a dict, turning
-
                 [config:mqtt]
                 host = 'localhost'
                 username = None
                 list = [1, 'aaa', 'bbb', 4]
-
             into
-
                 {u'username': None, u'host': 'localhost', u'list': [1, 'aaa', 'bbb', 4]}
-
             Cannot use config.items() because I want each value to be
             retrieved with g() as above '''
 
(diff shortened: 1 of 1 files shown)


jpmens:
Why is this done within `mqttwarn.py` and not within the service plugin? I don't like the case handling for a specific plugin within the main engine. Can you explain what you're trying to accomplish?

cuonglv1994:
I am trying to send the message from MQTTwarn to NSCA based purely on the payload (earlier I was told to do that based on the topics). With the changes I did, I could map the host name and service description from the payload to item['addr'] in processor(), which is later passed to the nsca service plugin. This way, I just need to config the topic and target once (like, [topic/mytopic/#] and targets= nsca: ), and don't have to remember which host I have just added to or removed from my network.
Well, but I forgot that I can do that by modifying a little bit in service plugins.
Thank you for your time, Sir.

PR #144, opened 2015-08-13
Update README.md

marcelrv:
Fixed typo in target and some help on how to use mqtt content when not using json formatting in the message


jpmens:
Merged, thanks!

Issue #157, opened 2015-12-01
HTTP enhancement to allow json POST


seblucas:
Hi,

First huge thanks for this project, it's really huge !!

Now the "issue" part, I just spent some time on using the http service to post data to firebase. Firebase need json in its POST payload and it's kind of hard to do with mqttwarn. The only reference I found is #71 and it seems too complicated for me. 

Would you accept a PR adding another parameter to specify if the query params should simply be jsonified (json.dumps) instead of only being urlencoded ?

Thanks in advance.

jpmens:
I would accept that gladly.

seblucas:
Closed, PR #158 was merged

PR #158, opened 2015-12-02
Added a fifth parameter to the HTTP service to send pure json

seblucas:
See #157 for more details.

I needed that to work with Firebase, I initially though of creating a plugin but using HTTP POST was enough for my case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -657,21 +657,22 @@ Requires:
 
 The `http` service allows GET and POST requests to an HTTP service.
 
-Each target has four parameters:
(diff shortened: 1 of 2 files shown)


seblucas:
One thing I forgot, if the Wiki is opened and you like my modifications to the README, I can also update the wiki as soon as this is merged.

jpmens:
Merged, thanks!

seblucas:
Thanks I updated the wiki page.

PR #160, opened 2015-12-08
Add a very basic Windows notification service.

mach327:
Spawns an icon and "balloontip" in the notification area of the Windows desktop.

Includes an MQTT icon from mqtt.org that is used for the notification area icon.
Includes two example screenshots of the end result while using - an ideal, and an example of what
happens if there are too many messages (lots of icons you can not read the message from).


sumnerboy12:
Hi Mach, thanks for your contribution. Unfortunately I am unable to test as all my servers are Debian based. 

However I do use the following for Windows/client notifications from mqttwarn (https://chrome.google.com/webstore/detail/mqtt2chrome/jfcmchhmjkddfoekjbkljcfpdjnpailh?hl=en) which I thought you might be interested in. It obviously requires you to run Chrome as your browser tho.

I personally would not want a notification service that required every notification to be clicked/mouse-overed, especially since I use mqttwarn extensively for home automation notifications!

mach327:
Yes, it's certainly not an ideal general Windows notification service. I wrote it as a quick way to get limited functionality that doesn't seem to otherwise exist without many heavy dependencies. So it meets my needs - others, maybe not so much, depending on what they're looking for.
Thanks for the tip towards mqtt2chrome, it looks like it would likely meet my needs - unfortunately I'm a Firefox user.

I was (still am, really) looking for an equivalent to [dunst (lightweight notification daemon)](https://github.com/knopwob/dunst) but on Windows. Writing this plugin looked like the quickest way to get the functionality I wanted most.

I don't know that it will enhance MQTTWarn to include `winnotify`, but it made sense to offer. :)

jpmens:
@mach327 You've gone to a lot of work with this, which I appreciate, and the idea is pretty good actually even though the implementation suffers from a few issues (I love the screenshot you made of the multiple notification icons ... :smile: )

If you think you can improve on that issue, I'll take this.

matbor:
@mach327 thx for the inspiration, just knocked up something in python that will display messages in the windows 10 notification center. https://github.com/matbor/mqtt2win10notify  Not sure if it will work on older versions of windows though.

balloon

jpmens:
Does that suffer from the same limitations as the original code? i.e. needs hover to dismis the icon?

mach327:
Currently, on Windows 7, yes. Actually, slightly worse, as the icons persist until the mqtt_notification process disappears. However, it seems to run properly in the background! :smile: I didn't realize there was a win32api module, but not only that - there's a "pythonnet" module! Basically the interface the powershell script was using, but in python, which looks like it should solve the can't-background issue.

I think I may have a solution for the hover-to-dismiss issue, too. NotifyIcon objects have a "Visible" property that seems to control whether the icon is in the taskbar. I'm testing now.

Issue #163, opened 2016-01-01
GSS Google Spreadsheet using disabled API, may need to use OAuth2


padelt:
I tried to use `services/gss.py` to add rows to a Google Spreadsheet from mqttwarn.
It seems that the [ClientLogin API](https://developers.google.com/api-client-library/java/google-api-java-client/client-login) used is both deprecated and disabled.

> Important: Do not use ClientLogin for new applications. Instead, use the more secure OAuth 2.0 authentication protocol. ClientLogin is a deprecated authentication protocol, which is being turned down on April 20, 2015. At that time, ClientLogin requests will no longer be answered.

It seems that using the OAuth 2.0 is now necessary.
https://developers.google.com/api-client-library/python/guide/aaa_overview

How to raise the problem:

All non-commented lines from `mqttwarn.ini`:
```
[defaults]
hostname     = 'localhost'  ; default
port         = 8883
username     = 'USERNAME'
password     = 'PASSWORD'
clientid     = 'mqttwarn'
lwt          = 'clients/mqttwarn'
skipretained = False
cleansession = False
protocol     = 3

ca_certs = '/volume/config/tls/ca.crt'
tls_insecure = True
tls_version = 'tlsv1'
logformat = '%(asctime)-15s %(levelname)-5s [%(module)s] %(message)s'
logfile   = '/volume/log/mqttwarn.log'
loglevel     = DEBUG
functions = 'samplefuncs'
launch   = file, log, gss

[config:file]
append_newline = True
targets = {
    'owntracks'  : ['/volume/log/mqtt-owntracks.log'],
    '_owntracks' : ['/volume/log/mqtt-owntracks-underscore.log'],
    'mqttwarn'   : ['/volume/log/mqttwarn.err'],
    }

[config:log]
targets = {
    'info'   : [ 'info' ],
    'warn'   : [ 'warn' ],
    'crit'   : [ 'crit' ],
    'error'  : [ 'error' ]
  }

[failover]
targets = log:error, file:mqttwarn

[hello/1]
targets = log:info
format = u'{name}: {number} => {_dthhmm}'

[owntracks-location]
topic = owntracks/+/+
targets = log:info, file:owntracks
datamap = OwnTracksTopic2Data()
format  = OwnTracksConvert()

[owntracks-location2]
topic = _owntracks/#
targets = log:info, file:owntracks

[owntracks-battery]
topic = owntracks/+/+
targets = log:info, file:owntracks
[...]

PR #164, opened 2016-01-02
gss2: Google Spreadsheet Service using OAuth2

padelt:
This is based on the `gss.py` service by @gumm but heavily modified to make it use OAuth 2.0 authentication so it actually works for me. Works on my Mac and a Docker Container; both running Python 2.7.9.
It should solve #163 .


jpmens:
Merged, thank you!

PR #168, opened 2016-02-04
Added postgres service based on mysql service

martyn-w:
Hi, this adds a postgres service, based on the same logic as the mysql service but adapted for postgres.

Please let me know if anything else is required in order to accept this PR?

Thanks,

Martyn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ _mqttwarn_ supports a number of services (listed alphabetically below):
 * [osxsay](#osxsay)
 * [pastebinpub](#pastebinpub)
 * [pipe](#pipe)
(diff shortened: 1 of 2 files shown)


jpmens:
Thank you for this, which looks rather good. Before accepting it: I need a bit for the README please, modeled after the other services therein.

martyn-w:
Hi @jpmens, I've updated the README with instructions for the postgres plugin.

Note that `full` is a reserved word in postgres, so I've called the fallback column `message` instead.

Best,

Martyn.

jpmens:
Excellent, thank you; Merged!

PR #172, opened 2016-02-22
Pushbullet - send notes to email/device/etc 

eschava:
Now next format for Pushbullet target is supported (recipient_type is optional):
'warnme' : [ 'API_KEY', 'receipient', 'recipient_type']
where recipient_type describes type of receipient
Could be one of:
device_iden (default)
email 
channel 
client


jpmens:
Merged, thanks.

Issue #177, opened 2016-04-09
Publish result of HTTP GET


thucke:
Hi,
I like to publish the result of an HTTP-GET command to the MQTT queue.
Example:
Having the configuration
    
    [config:http]
    timeout = 60
    targets = {
        #method     #URL               # query params or None          # list auth
        'dev16get'    : [ "get", "http://localhost:8083/ZWaveAPI/Run/devices[16].instances[0].Basic.data.level", None, (<username>, <password>) ],
    }
    
    [/home/zwave/Device16/switch/get]
    targets = http:dev16get

The command 

    mosquitto_pub -t /home/zwave/Device16/switch/get -n

should result in the following post to MQTT

    {"invalidateTime":1460221477,"updateTime":1460221478,"type":"int","value":0}

This would the the respone of the webserver if doing the same request in a web browser.
Is that possible?????

jpmens:
If I'm understanding you correctly it isn't possible. MQTTWARN listens to MQTT messages as a subscriber. Using the http service you can then issue a HTTP GET or POST request to an HTTP endpoint, which is the opposite of what you're asking.

thucke:
Thanks for this quick response.

I see this point.
On the other hand I wonder why MQTTWARN should support GET requests without being able featuring the response to modify the MQTT message.
Anyway, I find this would be a great feature of MQTTWARN.

BTW: MQTTWARN as it is this time is just a great enhancement - thank you so far!

jpmens:
What I think you are looking for is

```
HTTP -> MQTT
```

whereas _mqttwarn_ is the other way around. The reason we have a HTTP service which supports `GET` is in order to access HTTP endpoints which don't support `POST`.

What would definitely be possible is to have this:

```
MQTT -> MQTTWARN -> HTTP-GET -> MQTT publish
```

but, as it stands, this is not implemented (though relatively easy to implement I think). Although I also think for most this is overkill.

What you can attempt is to use a user-configurable function which is invoked on MQTT payloads, does the HTTP GET internally and modifies mqttwarn's notion of the payload before continuing. There are a couple of examples in the README

thucke:
Yeah - that's what I was thinking about if that feature would not be possible "out-of-the-box".
Creating a user-configurabe function requires to implement the HTTP GET mechanism on my own.
Hence I'd be glad using the "ready-for-use" HTTP GET mechanisms of MQTTWARN ;-)

[...]

PR #178, opened 2016-04-10
Periodic task improvements and useful demonstration

amotl:
Dear Jan-Piet and Ben,

find some improvements to the periodic task feature in this PR. We also added an example for using the feature in a way it could help @thucke with #177 as well as others wanting to achieve similar things.

Best,
Andreas.


jpmens:
Thank you very much: Merged!

PR #180, opened 2016-04-13
created sqlite_json2cols service

Vium:
It will create a column for each JSON entry and commit to the
configured sqlite db/table.


jpmens:
Very nice, thank you!

PR #184, opened 2016-04-19
New service serial

AnDann:
Send received data to serialport.
Data from serialport will be ignored.
I use it as mqtt to serial bridge in my openhab environment


jpmens:
I will take this with thanks.

BTW, in future, when you clean up files outside of your main contribution it's better to provide separate pull requests.

PR #188, opened 2016-05-05
Optional LWT; fix for pyserial with isOpen

mortenf:
Hi,

My version of pyserial (2.6 with python 2.7.9) doesn't have is_open, as required by the serial module. It does however have isOpen.

Also, I made LWT optional, by setting the topic in the configuration file to "None".

### DIFF
--- a/README.md
+++ b/README.md
@@ -825,15 +825,19 @@ You should first create an application and respective event following the [tutor
 
 Afterward you will find your Application ID and Application Secret in the "Basic Info" of your application.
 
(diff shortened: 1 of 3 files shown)


mortenf:
Added a change to the Instapush documentation, to make it clear(er) how to use it with payload based tracker values (per #121).

jpmens:
Merged, thank you.

If you don't mind, I'd prefer individual PRs for changes to different things in future; I find this makes it easier to judge / check changes.

mortenf:
Thank you, and agreed - I'll make sure to split into separate PRs (even if it's extra work with extra branches).

PR #170, opened 2016-02-16
Use functions for computing topic targets

amotl:
Good morning Jan-Piet,

we have another improvement on top of the "dynamic topic targets" feature #169. With this patch, _mqttwarn_ users will be able to compute topic targets by python functions, in the same style of the `datamap` / `alldata` machinery.

@einsiedlerkrebs already is working on the documentation part to outline an advanced scenario. This will take some time though.

Best,
Andreas.

PR #169, opened 2016-02-14
Add feature "dynamic topic targets"

amotl:
Dear Jan-Piet and Ben,

first things first: Thanks for sharing this great project!

We tried to enhance _mqttwarn_ to add functionality we need for our Hiveeyes bee monitoring project. The feature was implemented as generic as possible to provide added value to the _mqttwarn_ infrastructure in general. Kudos go to @einsiedlerkrebs for conceiving it and making this possible.

A short overview over the changes made:
- Interpolate transformation data into topic targets
- Improve overall robustness and address #27, this was essential for implementing the feature

Please get back to us if you have further questions or suggestions.

With kind regards,
Andreas.

PR #192, opened 2016-05-30
Various improvements

amotl:
Dear Jan-Piet and Ben,

we have another round of improvements to *mqttwarn* ready. This is the foundation of the implementation of our prototypic Hiveeyes Schwarmalarm #193 on top of *mqttwarn*, which is the preliminary alerting component of the [Hiveeyes system]. We are building upon one of our former contributions, see also [Incorporate topic names into topic targets] and #169, #170.

The changes are mainly about overall robustness, especially when having more pressure on the job queue. The shutdown behavior of *mqttwarn* hasn't been optimal before, we often had cases where *mqttwarn* wouldn't stop at all on `SIGTERM` because the worker (`def processor`) already had been shut down while `def cleanup` was waiting for the notification job queue to drain, resulting in deadlock situations only to be mitigated by `kill -9` on our systems.

Along the lines, we added and improved some log messages here and there, integrated some minor features not worth a separate pull request in our opinion and also added a conveniency routine to monkeypatch the xmpppy library code for mitigating a known and popular problem when connecting to XMPP servers with TLS support (can be used optionally, thus non-invasive).

The details:
- Make “num_workers” setting configurable
- Add possibility to send log messages to stderr
- Improve reporting about log level
- Log warning message if transform_data cannot be formatted properly
- Improve worker robustness and shutdown behavior re. job queue draining
- Improve logging and robustness for service plugins smtp and xmpp
- Add monkey patch for xmpp service plugin to mitigate problem with SSLSocket X.509 issuer

Please let us know what you think about this. In the meanwhile, we will prepare the main PR mentioned above. Thanks in advance for your efforts!

Cheers, Andreas.

[Hiveeyes system]: https://hiveeyes.org/docs/system/
[Incorporate topic names into topic targets]: https://github.com/jpmens/mqttwarn/wiki/Incorporating-topic-names#incorporate-topic-names-into-topic-targets

### DIFF
--- a/services/smtp.py
+++ b/services/smtp.py
@@ -9,6 +9,7 @@ import smtplib
 from email.mime.text import MIMEText
 
 def plugin(srv, item):
+    """Send a message to SMTP recipient(s)."""
 
     srv.logging.debug("*** MODULE=%s: service=%s, target=%s", __file__, item.service, item.target)
 
@@ -26,8 +27,13 @@ def plugin(srv, item):
     msg['From']         = sender
     msg['X-Mailer']     = srv.SCRIPTNAME
 
+    if not smtp_addresses:
+        srv.logging.warn("Skipped sending SMTP notification to %s, "
+                         "no addresses configured: %s" % (item.target, smtp_addresses))
+        return False
+
     try:
-        srv.logging.debug("Sending SMTP notification to %s [%s]..." % (item.target, smtp_addresses))
+        srv.logging.debug("Sending SMTP notification to %s, addresses: %s" % (item.target, smtp_addresses))
         server = smtplib.SMTP(server)
         server.set_debuglevel(0)
         server.ehlo()
@@ -39,7 +45,8 @@ def plugin(srv, item):
         server.quit()
         srv.logging.debug("Successfully sent SMTP notification")
     except Exception, e:
-        srv.logging.warn("Error sending notification to SMTP recipient %s [%s]: %s" % (item.target, smtp_addresses, str(e)))
+        srv.logging.warn("Error sending notification to SMTP recipient %s, addresses: %s. "
+                         "Exception: %s" % (item.target, smtp_addresses, str(e)))
         return False
 
     return True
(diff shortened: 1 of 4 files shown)


services/smtp.py
```diff
@@ -26,8 +27,13 @@ def plugin(srv, item):
     msg['From']         = sender
     msg['X-Mailer']     = srv.SCRIPTNAME
 
+    if not smtp_addresses:
+        srv.logging.warn("Skipped sending SMTP notification to %s, "
+                         "no addresses configured: %s" % (item.target, smtp_addresses))
```
sumnerboy12: Isn't the second parameter here redundant?

services/xmpp.py
```diff
@@ -17,8 +17,13 @@ def plugin(srv, item):
     password = item.config['password']
     text = item.message
 
+    if not xmpp_addresses:
+        srv.logging.warn("Skipped sending XMPP notification to %s, "
+                         "no addresses configured: %s" % (item.target, xmpp_addresses))
```
sumnerboy12: Same here?

sumnerboy12:
Looks really good Andreas - thank you! I will leave JP to comment further but I took a good look around your project and am mightily impressed - I can imagine this will be a very useful reference project for others want to setup similar monitoring applications using MQTT/mqttwarn. Thanks again for sharing your code with our little project!

services/smtp.py
jpmens: Yes, because it'll be `None`. I will remove after merge

jpmens:
@amotl this looks excellent thank you. And thank you *particularly* for finally finding and fixing the suboptimal shutdown behavior. ;-)

jpmens:
Merged!

services/smtp.py
amotl: Sure, the second parameter will be None or an empty list. Thanks for spotting this!

services/xmpp.py
amotl: Right. :-)

amotl:
Thanks!

PR #194, opened 2016-05-31
Example data loss sensor with notifications based on periodic tasks

amotl:
Dear @jpmens and @sumnerboy12,

thanks again for merging our little example for mqttwarn so quickly. While we have to move on to other features still lingering, we added another sensor to the Hiveeyes mqttwarn subsystem, which is important  to us: We also want to get notified on data loss events, i.e. if a sensor node suddenly stops sending data through any cause, as a matter of fact mostly originating from drained batteries.

The behavior of the implementation of the added `hiveeyes_dataloss_monitor` tries to reach DWIM, i.e.:
- It will just fire a notification once and not each and every time when checking against the configured data loss timeout during each check interval
- It also does not have to be configured in any significant way as it automatically watches all sensor nodes ever seen on the bus
 
This time, just a minor improvement to *mqttwarn* has been made: The `Service` object now enables plugins to use *mqttwarn*'s internal infrastructure (don't tell anybody! ;]). The `hiveeyes_dataloss_monitor` is now able to reach out to *mqttwarn*'s `send_to_targets` routine for actually sending data loss notifications to our beekeepers.

Cheers,
Andreas on behalf of the Hiveeyes Developers.

### DIFF
--- a/mqttwarn.py
+++ b/mqttwarn.py
@@ -358,9 +358,19 @@ ptlist = {}         # List of PeriodicThread() objects
 # and its global instantiation
 class Service(object):
     def __init__(self, mqttc, logging):
+
+        # Reference to MQTT client object
         self.mqttc    = mqttc
+
+        # Reference to all mqttwarn globals, for using its machinery from plugins
+        self.mwcore   = globals()
+
+        # Reference to logging object
         self.logging  = logging
+
+        # Name of self ("mqttwarn", mostly)
         self.SCRIPTNAME = SCRIPTNAME
+
 srv = Service(None, None)
 
 service_plugins = {}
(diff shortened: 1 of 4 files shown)


jpmens:
Clicking on the MERGE button for you is turning into real work, @amotl !!!

Thank you.

PS: I don't just click the button; I also actually test whether `mqttwarn` at least starts up. 😆

mqttwarn.py
```diff
@@ -358,9 +358,19 @@ def join(self):
 # and its global instantiation
 class Service(object):
     def __init__(self, mqttc, logging):
+
+        # Reference to MQTT client object
         self.mqttc    = mqttc
+
+        # Reference to all mqttwarn globals, for using its machinery from plugins
+        self.mwcore   = globals()
```
jpmens: What could possibly go wrong ... :-) 
Good idea!

amotl:
Wow, that was quick. Thanks again!

jpmens:
I couldn't find you on Twitter, so I had to give credit this way; I hope you don't mind.

[https://twitter.com/jpmens/status/737679714651869185](https://twitter.com/jpmens/status/737679714651869185)

amotl:
Appreciate it! Thanks, @jpmens! 👍

PR #195, opened 2016-06-02
add support for ssh

DavidVentura:


jpmens:
Interesting.. Can you say how you envision this being used?

That said, I'll take this, but please make user/password optional so that the connection defaults to using a (specified) SSH key file, the passphrase for which should come from SSH agent so that we don't have that in a configuration file either.

DavidVentura:
My primary idea was to get remote notifications on desktop (like linuxnotify / dbus services, but remotely).

I have opened an issue with paramiko regarding this. The library seems to be ignoring the ssh config file (Or I forgot to do something with it). Once it's resolved I'll update with it.

Anyway I thought this might be accepted given (for example) the smtp user/password is on the config file

jpmens:
You're right about many of the _targets_ which have credentials in the config file because there's simply no clever alternative. :)

The issue I have with SSH is that this service could also be used to login to the remote server, thereby possibly causing havoc.

I'll sleep it over and consider it.

In any case, thank you for your contribution; much appreciated!

DavidVentura:
Can I get `item.message` as an object instead of a string? in that case I could run a token replacement over fixed commands instead of arbitrary commands:

```
targets = {
    's01'    : [ 'DISPLAY=:0 notify-send %s' ]
    }
```

jpmens:
If item.message contains, say, a JSON string, you can convert it to an object with json.loads()

DavidVentura:
YOu've responded to my old message, which I've edited. I'm letting you know

On 2 June 2016 at 15:36, JP Mens <notifications@github.com> wrote:

> If item.message contains, say, a JSON string, you can convert it to an
> object with json.loads()
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/jpmens/mqttwarn/pull/195#issuecomment-223382437>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ADe0boZBI_oLcRtmwKbXUQ8O35R32hF0ks5qHyK_gaJpZM4IsvzM>
> .
>



-- 
*Stack* is the new term for "I have no idea what I'm actually using".

jpmens:
(One of the problems when responding via mail...)

jpmens:
@DavidVentura you've put a lot of effort into this, for which you have my thanks. I've decided to take it. However, for reasons I find difficult to word, this module frightens me. 😄 As such I will add a short note of warning to your portion of the README.

jpmens:
Merged.

PR #196, opened 2016-06-02
Add systemd unit configuration file

amotl:
Dear Jan-Piet and Ben,

we installed *mqttwarn* as a system daemon on the Hiveeyes data collection server, please find the systemd unit and logrotate configuration files in this PR. They could be useful to others as well.

Using a Python virtualenv for isolating the installation from the system Python reflects the way we are currently running *mqttwarn* on our machines, adapting this to other flavors by amending the `WorkingDirectory ` and `ExecStart` directives in `mqttwarn.service` should be easy for everyone to do.

With kind regards,
Andreas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2392,6 +2392,11 @@ You'll need at least the following components:
 
 I recommend you use [Supervisor](http://jpmens.net/2014/02/13/in-my-toolbox-supervisord/) for running this.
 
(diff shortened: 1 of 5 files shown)


jpmens:
Brilliant! Merged with thanks, Andreas.

amotl:
Thanks again!

PR #198, opened 2016-06-04
Accept key pair defined in config file instead of user/password

DavidVentura:


jpmens:
You say "Accept key pair defined in config file"; it's not clear to me (README missing) where in mqttwarn's configuration you specify which key to use.

DavidVentura:
You are right. I'll update the README shortly. The idea is that if username and password are both missing from config, the program will read user's ssh config file and load from there:

* Username, keyfile (obligatory)
* Port, hostname (optional)

DavidVentura:
I've updated the README accordingly. Let me know what you think

jpmens:
Merged, thank you.

PR #201, opened 2016-06-27
fix thingspeak issue when fields are received in the wrong order

marcelrv:
this fixes thingspeak issue when builddate is used and the fields are received in the wrong order


jpmens:
Merged, thank you

PR #202, opened 2016-07-07
Configurable zabbix host with MQTT host discovery template attached

marianblazicek:
I've spent few hours debugging why my zabbix won't create new host from discovery. Pulling out the host name to configuration (.ini) have solved it more elegantly then creating dummy host MQTT_BUS.


jpmens:
Merged, thanks.

PR #203, opened 2016-07-14
Support html messages in smtp service

qk4l:
Hi I added html message sending for `smtp` service.
HTML mails contain two path `plain` and `html` and have Content-Type: multipart/alternative so MUAs that do not support html can read mail as plain text.

This feature is optional and disable by default.


jpmens:
Merged, thank you.

PR #209, opened 2016-09-24
add execute service

tobru:
This commit adds a new service to simply execute a command
when somehing is published. The data of the published message
can be passed as commandline argument if needed.


jpmens:
Merged. Thank you!

PR #210, opened 2016-10-15
Add as_user option to slack service

mortenf:
To allow for better control of message appearance in Slack, I have added the 'as_user' option from the Slack API to the target configuration.

The default value is False, which is also the current (intended) behaviour.


jpmens:
Merged, thanks!

PR #212, opened 2016-10-19
Fix partial serial fix for is_open

mortenf:
Due to what must be an error on my part, parts of the serial service fix for `is_open` in #188 is missing from the current branch.

This should correct the omissions.


jpmens:
Merged, thanks!

PR #213, opened 2016-10-22
Add item data interpolation to thingspeak service

mortenf:
To make it possible to send to ThingSpeak with a JSON-formatted message containing multiple values, I have extended the service target definition. It is now possible to define a target using an ordered list of item data field names, the values of which will be assigned to the ThingSpeak fields.

The original usage pattern is still in place.


jpmens:
Merged, thanks!

PR #214, opened 2016-10-26
Add service aliasing with module option

mortenf:
Sometimes it'd be nice (or necessary) to have two versions of a specific service, e.g. for notifying two different MySQL servers or to have one file appended to and another one truncated.

For this reason I have added a non-mandatory option to service configurations, to essentially allow aliasing of service names as used in `launch` and `targets` options and in `config:xxx` sections: The option `module` in a `config:xxx` section names the real service (module file) to use, allowing for both of e.g. `file` and `appendfile`, the last of which then needs to have `module = file`.

I have checked all current services, none of which use an option called `module`. Also, the option is excluded from the service configuration dictionary, like `targets`.

The README is updated, although I'm sure it could be improved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ The `functions` option specifies the path to a Python file containing functions 
 
 ### `launch`
 
-In the `launch` option you specify which _services_ (of those available in the `services/` directory of _mqttwarn_) you want to be able to use in target definitions.
+In the `launch` option you specify which _services_ (of those available in the `services/` directory of _mqttwarn_ or using the `module` option, see the following paragraphs) you want to be able to use in target definitions. 
 
 ## The `[config:xxx]` sections
 
@@ -200,7 +200,8 @@ has a mandatory option called `targets`, which is a dictionary of target names, 
 pointing to an array of "addresses". Address formats depend on the particular service.
 
 A service section may have an option called `module`, which refers to the name
-of the actual service module to use. As such, a service called `filetruncate`
+of the actual service module to use. A service called `filetruncate` - and 
+referenced as such in the `launch` option -
 may have `module = file`, in which case the service works like a regular `file`
 service, with its own distinct set of service options. It is thus possible to
 have several different service configurations for the same underlying service,
--- a/README.md
+++ b/README.md
@@ -199,6 +199,14 @@ Sections called `[config:xxx]` configure settings for a service _xxx_. Each of t
 has a mandatory option called `targets`, which is a dictionary of target names, each
 pointing to an array of "addresses". Address formats depend on the particular service.
 
+A service section may have an option called `module`, which refers to the name
+of the actual service module to use. As such, a service called `filetruncate`
+may have `module = file`, in which case the service works like a regular `file`
+service, with its own distinct set of service options. It is thus possible to
+have several different service configurations for the same underlying service,
+with different configurations, e.g. one for files that should have notifications
+appended, and one for files that should get truncated before writes.
+
 ## The `[failover]` section
 
 There is a special section (optional) for defining a target (or targets) for internal error conditions. Currently there is only one error handled by this logic, broker disconnection.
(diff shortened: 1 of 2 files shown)


jpmens:
This sounds very interesting. Looking at your modified README it looks to me as though something's missing, as you reference a section called `#the-configxxx-sections` which doesn't exist. Would you double-check please, and if necessary add that as a commit to this PR?

mortenf:
That anchor is auto generated by github from the headline "The [config:xxx] sections" when the README is rendered. You can see it in action here: https://github.com/jpmens/mqttwarn/#the-configxxx-sections

When hovering over a header on github, a chain/link icon is shown to the left of the header. The icon contains a link to the header, including the anchor.

jpmens:
So, for an existing service `log` such as

```ini
[config:log]
targets = {
    'debug'  : [ 'debug' ],
    'info'   : [ 'info' ],
    'warn'   : [ 'warn' ],
    'crit'   : [ 'crit' ],
    'error'  : [ 'error' ]
  }
```

I should be able to create a second service called `xlog` like this?

```ini
[config:xlog]
module = log
targets = {
    'debug'  : [ 'debug' ],
  }
```

and use this for topic `mens/1` as 

```ini
[mens/1]
targets = xlog:debug
```

? I can't get this to work: mqttwarn says 

```
ERROR [mqttwarn] Invalid configuration: topic mens/1 points to non-existing service xlog
```

mortenf:
Thanks for taking the time to test this!

I'm guessing you haven't added the `xlog` service to `launch`?

This should of course be spelled out in the README - hopefully it should be clearer now.

jpmens:
I did actually, but the process doesn't start at all then (understandably?)

```ini
launch = log, xlog
```

```
ERROR [mqttwarn] Can't load xlog service (services/xlog.py): [Errno 2] No such file or directory: u'services/xlog.py'
```

jpmens:
Trimming it all down to a trivial config like this:

```ini
[defaults]
hostname  = 'localhost'  ; default
port      = 1883
lwt       = 'clients/mqttwarn'
skipretained = True
cleansession = True
clientid = mqttwarn-2-tiggr2
protocol     = 4

logformat = '%(asctime)-15s %(levelname)-5s [%(module)s] %(message)s'
logfile   = 'mqttwarn.log'
loglevel     = DEBUG
launch	 = file, log, xlog

[config:log]
targets = {
    'debug'  : [ 'debug' ],
    'info'   : [ 'info' ],
    'warn'   : [ 'warn' ],
    'crit'   : [ 'crit' ],
    'error'  : [ 'error' ]
  }

[config:xlog]
module = log
targets = {
    'debug'  : [ 'debug' ],
  }


[mens/1]
targets = xlog:debug
```

Launching mqttwarn produces this stack (line 966 doesn't protect service from not having `module`)

```
Traceback (most recent call last):
  File "./mqttwarn.py", line 1099, in <module>
    connect()
  File "./mqttwarn.py", line 993, in connect
    load_services(services)
  File "./mqttwarn.py", line 966, in load_services
    module = service_config.get('module', service)
AttributeError: 'NoneType' object has no attribute 'get'
```

If I'm doing something wrong, please show me a minimal config which works.

jpmens:
That fixes it. Thanks for a) doing this, b) having patience with me. I will add a bit of an example to your description and then we're done I think.

mortenf:
There was a problem with the use of the `cf.config` method in `load_services`, as it was wrapped in a `try`/`catch`, but the method doesn't throw an exception if a section is not found, just `None`. This had me puzzled for a while and accounts for half of the problem.

As for the rest of the problem, I had my tests mixed up, so I didn't catch that the removal of the `module` option from the service configuration also affected the load process...

Both of the above should now be fixed - trying out your trivial configuration above works (after removing `file` from `launch`).

Sorry about the failed tests.

mortenf:
Thanks! Examples are always good, indeed.

PR #216, opened 2016-10-31
Set JSON content type for POST in http service

mortenf:
Services like IFTTT insist that POST requests carry an HTTP header indicating the content type, e.g. for JSON:
`Content-Type: application/json`

This change adds that header.

Previously there was no indication of content type, so the change should hopefully not affect anyone negatively.


jpmens:
Sensible. Thank you: merged.

PR #223, opened 2016-12-14
support for different slack accounts (non-breaking changes)

sumnerboy12:


sumnerboy12:
@jpmens could you please review this? need to account for two optional target parameters now (token and as_user).

jpmens:
Merged, thanks Ben

PR #229, opened 2016-12-24
Add TLSv1.1 and TLSv1.2 as available config options

coffeybeans:
Added capability to use TLSv1.1 and TLSv1.2 as config options.
My first PR so any feedback is welcome.


jpmens:
Merged, thank you.

PR #232, opened 2017-02-06
Support using templates with influxdb service

schinken:
Payload seems to always contain the original value, even after "templating", so we should use item.message here


nickbalch:
+1 for this pull ... I've edited the relevant file manually for now, but it would be great to have this merged

jpmens:
Merged, thank you.

PR #244, opened 2017-05-28
Add celery plugin

orhanhenrik:
Celery is a python task queue. This plugin allows mqttwarn to send messages from topics to the task queue so workers can consume them. This can be useful because many people already use celery and can simply reuse the same queue to process messages from MQTT.

Please let me know if I should add more docs to the code or readme.


jpmens:
This looks very useful; thank you. Merged!

PR #246, opened 2017-06-06
Allow manually setting chatted for telegram service

MrBramme:
added the option to specify the chatid for the telegram service manually. a warning, this will need to be the case for all targets in this notifier as the "use_chat_id" bool is defined once for all targets in it's config.


jpmens:
Merged, thanks.

PR #249, opened 2017-06-21
Support urls in osxnotify

afrittoli:
When using osx desktop notification, it can be useful to have a
URL associated to the notification, so that it may be clicked to
open a web page with further details.

I'm using this with Firehose from OpenStack when listening to gerrit
topics, to open the corresponding change in gerrit. The url can be
set on the item using payload transformation functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1519,9 +1519,11 @@ Requires:
 
 * Requires Mac ;-) and [pync](https://github.com/setem/pync) which uses the binary [terminal-notifier](https://github.com/alloy/terminal-notifier) created by Eloy Durán. Note: upon first launch, `pync` will download and extract `https://github.com/downloads/alloy/terminal-notifier/terminal-notifier_1.4.2.zip` into a directory `vendor/`.
 
-| Topic option  |  M/O   | Description                            |
-| ------------- | :----: | -------------------------------------- |
-| `title`       |   O    | application title (dflt: topic name)   |
+| Topic option  |  M/O   | Description                                     |
+| ------------- | :----: | ----------------------------------------------- |
+| `title`       |   O    | application title (dflt: topic name)            |
+
+It `url` is defined in items.data, it's value is passed to the notification, so that the URL is opened when the notification is clicked.
 
 ![osxnotify](assets/osxnotify.jpg)
 
--- a/services/osxnotify.py
+++ b/services/osxnotify.py
@@ -15,8 +15,14 @@ def plugin(srv, item):
     text = item.message
     application_name = item.get('title', item.topic)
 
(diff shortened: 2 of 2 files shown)


afrittoli:
The data from the transformation functions ends up in item['data'], so this is not really useful.
Items that are set in the config file end up in item['config'], which is not used either. 
So this will need some more work.

jpmens:
Thanks for your contribution. I was just looking at this and came to the same conclusion. Go ahead and fix and just push further commits into this: I'll keep it open.

afrittoli:
I updated the pull request to get url from item.data which can be set using a transformation function e.g. 

[__topic__]
alldata = all_data()


def all_data(topic, data, srv):
    extra_data = json.loads(data['payload'])
    # Make the change data (including url) available for formatting
    return extra_data['change']

jpmens:
The code works, but I don't see any sort of URL or possibility of clicking in the notification. I have verified that `url=` is set in the service

jmbp-3643

jpmens:
This is a bit of a mess, though probably not your fault.

The release of terminal-notifier installed by _pync_ works:

```
./pync/pync/vendor/terminal-notifier-1.6.1/terminal-notifier.app/Contents/MacOS/terminal-notifier -open http://localhost/blurb -message kdkk
```

I can click on the bubble and Safari opens with the URL.

But that's not what pync does:

```python
from pync import Notifier

Notifier.notify('Hello World', open='http://github.com/')
```

afrittoli:
My configuration for reference:

```
[defaults]
hostname     = 'firehose.openstack.org'
port         = 1883
username     = None
password     = None
clientid     = 'mqttwarn'
lwt          = 'clients/mqttwarn'
skipretained = False
cleansession = False

protocol     = 3

; logging
logformat = '%(asctime)-15s %(levelname)-5s [%(module)s] %(message)s'
logfile   = 'mqttwarn.log'

loglevel  = WARN

launch    = osxnotify

functions = osx_notify_format

[config:osxnotify]
targets = {'desktop': ''}

[gerrit/openstack/tempest/comment-added]
alldata = all_data()
title = "{project}"
format= "{type} on {number}:\n{comment:.100}"
targets = osxnotify:desktop
```

I'm not sure I get what the problem is - I tried with pync directly and it works fine for me.

```
$ pip freeze | grep pync
pync==1.6.1
$ python
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 12:39:47) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
```

jpmens:
I had the wrong version. Merging, thank you.

PR #250, opened 2017-06-22
rrdtool service amendment + documentation

otdftr:
combined my two patches into one... didn't check for the correct repo /branch before


jpmens:
Merged, thank you.

PR #254, opened 2017-07-10
Added hipchat service

lrizzi:
This is the first version of the hipchat plugin for room notifications, both cloud and hosted versions are tested and supported.


jpmens:
Merged, thank you, and thanks for the good documentation.

PR #256, opened 2017-07-17
Fix function loading

sumnerboy12:
@jpmens this is a breaking change unfortunately but was necessary to allow specifying an absolute path for a `functions` file. The reason for this was I wanted to move my `mqttwarn` config into a separate git repo for version management etc, this included `functions.py`. However specifying the full path to the new `functions.py` location in `mqttwarn.ini` was not working.

The change in this PR allows for any path to be specified for `cf.functions`, but the `.py` extension must be included - i.e. it must be the full filename.

Have a look and see what you think. Understand if you don't want to merge, if so I will just run this locally.

### DIFF
--- a/mqttwarn.ini.sample
+++ b/mqttwarn.ini.sample
@@ -27,8 +27,7 @@ logfile   = 'mqttwarn.log'
 loglevel  = DEBUG
 
 ; path to file containing self-defined functions for formatmap and datamap
-; omit the '.py' extension
-functions = 'samplefuncs'
+functions = 'samplefuncs.py'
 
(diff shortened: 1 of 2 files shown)


jpmens:
Merged thanks. To be honest: I've always detested this comment: `; omit the '.py' extension` 😄

elmicha:
Hi,

the README.md still says `; omit the '.py' extension`. I always got:

    WARNING [mqttwarn] Cannot invoke filter function owntracks_filter defined in test/payload: local variable 'py_mod' referenced before assignment

But with extension it all works! Thank you for the great program!

PR #260, opened 2017-08-05
Boxcar

psyciknz:
Hi

I've been speaking to @sumnerboy12 regarding mqttwarn.  I was a big proponent of boxcar to ios notifications, and use them a bit on my OpenHAB solutions.  So I've added boxcar support.  But It may have gone from the app store so it may negate it use.  Is there something similar you already run?  Pushover?  

Boxcar does ios (and android I think) notifications, where you can set the warning sound on the message, and the text can contain links etc.  http://blog.boxcar.io/post/93211745502/boxcar-api-update-boxcar-api-update-icon-and

Thoughts?

### DIFF
--- a/README.md
+++ b/README.md
@@ -742,7 +742,7 @@ Requires
 
 The `boxcar` service will use the boxcar notification service to send messages.  It was one of the first IOS notification services..
 
-The paylod (message) of the mqtt topic, can use the "|" seperator to contain a notification title, a long message and a url.  
+The payload  of the MQTT message, can use the "|" seperator to contain a notification title, a long message and a url.  
 Urls can go in the long message with an <a href> tag.  Most other html tags will work also.
 To get the message to open a url, use the 3rd section for a url.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -737,6 +738,73 @@ Requires
 * [Asterisk](http://www.asterisk.org/) with configured AMI interface (manager.conf)
 * pyst2 -  powerful Python abstraction of the various Asterisk APIs (pip install pyst2)
 
+## `boxcar`
+
+The `boxcar` service will use the boxcar notification service to send messages.  It was one of the first IOS notification services..
+
+The paylod (message) of the mqtt topic, can use the "|" seperator to contain a notification title, a long message and a url.
```
jpmens: an mqtt topic doesn't have a payload. Please rephrase and it's spelled "MQTT"

psyciknz: Is it worth continuting with this PR given that Boxcar is: A not multiplatform (IOS only) and B, has been pulled from the IOS app store.

I didn't realise this before making the changes.

jpmens:
Ah, now I understand: it's completely gone. OK, you're right: thanks for your work (and I'm sorry you went through it), but I won't take this now. If you leave your fork alive, people will find this there.

PR #266, opened 2017-09-28
Allow user to specify postgres schema

clarkspark:
My understanding is that postgres uses the 'public' schema by default.  This change will continue to use the public schema when the user does not specified a schema or use the user specified schema instead.  I updated the README.md as well.


jpmens:
Merged, thank you.

PR #267, opened 2017-10-06
New version of fbchat, some needed adjustments

przemas75:
[fbchat](https://github.com/carpedm20/fbchat) has bee updated, hence this plugin needs adjustments. Documentation is the same.


jpmens:
Merged thanks.

Issue #280, opened 2017-12-25
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3


semaf:
I have used in my Text "Wohnungstür" and it seems that the "ü" is a problem.

```
Traceback (most recent call last):
  File "mqttwarn.py", line 1122, in <module>
    connect()
  File "mqttwarn.py", line 1062, in connect
    mqttc.loop_forever()
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 1481, in loop_forever
    rc = self.loop(timeout, max_packets)
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 1003, in loop
    rc = self.loop_read(max_packets)
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 1284, in loop_read
    rc = self._packet_read()
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 1849, in _packet_read
    rc = self._packet_handle()
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 2305, in _packet_handle
    return self._handle_publish()
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 2500, in _handle_publish
    self._handle_on_message(message)
  File "/usr/local/lib/python2.7/dist-packages/paho/mqtt/client.py", line 2647, in _handle_on_message
    self.on_message(self, self._userdata, message)
  File "mqttwarn.py", line 609, in on_message
    logging.debug("Message received on %s: %s" % (topic, payload))
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 9: ordinal not in range(128)
```

amotl:
Bummer! We will probably have to get better regarding Unicode awareness.

I just gave this a quick try by doing
`echo '{"name": "Wohnungstür", "number": 42}' | mosquitto_pub -h localhost -t hello/1 -l`

while running `mqttwarn.py` using the `mqttwarn.ini.sample` and it properly said
`2017-12-25 14:12:32,551 DEBUG [mqttwarn] Message received on hello/1: {"name": "Wohnungstür", "number": 42}`.

Can you tell us what you did exactly to be able to reproduce this? Please also tell us which _Python_ and _paho-mqtt_ versions you are using and whether you use _Mosquitto_ or any other MQTT broker. Thanks!

semaf:
[...]

Issue #276, opened 2017-12-01
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe7 in position 31: invalid continuation byte


knsg0d:
Traceback (most recent call last):
File "test123.py", line 24, in
client.loop_forever()
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 1481, in loop_forever
rc = self.loop(timeout, max_packets)
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 1003, in loop
rc = self.loop_read(max_packets)
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 1284, in loop_read
rc = self._packet_read()
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 1849, in _packet_read
rc = self._packet_handle()
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 2305, in _packet_handle
return self._handle_publish()
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 2500, in _handle_publish
self._handle_on_message(message)
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 2647, in _handle_on_message
self.on_message(self, self._userdata, message)
File "test123.py", line 12, in on_message
print(msg.topic+" "+str(msg.payload))
File "/usr/local/lib/python2.7/dist-packages/paho_mqtt-1.3.1-py2.7.egg/paho/mqtt/client.py", line 360, in topic
return self._topic.decode('utf-8')
File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xe7 in position 31: invalid continuation byte

jpmens:
What exactly do you expect me to do with that traceback?

jpmens:
Closed due to inactivity.

PR #281, opened 2017-12-25
Try to decode message payload from UTF-8 before further processing

amotl:
Hello again Jan-Piet and Ben,

we tried to help @semaf with this PR to address #280 and maybe also #276. The issue was that _mqttwarn_ croaked and even died when receiving MQTT payloads containing things in utf-8 character encoding, like "Wohnungstür".

With kind regards,
Andreas.


jpmens:
Merged, thank you!

Issue #146, opened 2015-09-01
Unwanted 'u' character appearing in JSON name and value fields processed by 'format'


rapport:
When I publish humidity data to my MQTT broker...

``` python
mosquitto_pub -h 192.168.178.20 -t test/input -m '{"d":{"humidity":"62.18"}}'
```

And, then try to process with the mqttwarn mqtt service I appear to get unwanted 'u' characters in front of all the name and values which are nested inside the JSON that are processing with the default format processing  (as seen test/output below). e.g I get 'u' appearing like this

``` python
test/output "{u'humidity': u'62.18'}"
```

My mqttwarn.ini config contains the following settings:

``` python
[config:mqtt]
  hostname =  'localhost'
  port =  1883
  qos =  0
  retain =  False
  username =  None
  password =  None
  targets = {
    'o1'    : [ 'test/output' ],
  }
[test/input]
  targets = mqtt:o1
  format = "{d}"
```

I assume it is a bug that the 'u' appears in this context! It looks like it is being inserted by some processing of quotations as when publish without a quotation first instance on the 62.18 data field e.g.

``` python
mosquitto_pub -h 192.168.178.20 -t test/input -m '{"d":{"humidity":62.18}}'
```

I get no chartacter appearing in front of the 62.18 value e.g.

``` python
test/output "{u'humidity': 62.18}"
```

I am able to work around the problem by calling a custom function from format.  i.e.

``` python
   format = CustomTextTransform()
```

functions.py

``` python
def CustomTextTransform(message):
    data = json.dumps(message['d'])
    return str(data)
```

rapport:
I just noticed when publishing value without quotations there is no extra character appearing

mosquitto_pub -h 192.168.178.20 -t test/input -m '{"d":{"humidity":62.18}}'

test/output "{u'humidity': 62.18}"

rapport:
Seems to be a bit similar to this problem
http://stackoverflow.com/questions/12966308/decoding-json-with-python-and-storing-nested-object

jpmens:
Please provide a reproducible example with your configuration and the published payload. (Kindly use Github's markup for formatting code & examples :-)

rapport:
I have reformatted previous comment and hopefully the example is clearer now!

jpmens:
Thank you for your concise report. It looks strange and honestly I don't know at this moment what to do about it ... I think your use of a custom function is currently the only solution. I will look into this a bit closer when I get an opportunity to.

[...]

PR #282, opened 2017-12-25
Optional JSON output serialization using type coercion before formatting

amotl:
Dear Jan-Piet and Gary (@rapport),

we just found #146 and tried to solve it without using a custom function. The problem was that _mqttwarn_ puts things like `{u'humidity': u'62.18'}` on its egress data path, which is the String-serialized **Python representation** of the respective data structure.

Receivers **expecting JSON** will probably croak on that, so @rapport worked around it by implementing a custom function which does the data extraction and JSON serialization manually.

With this PR, it will become possible to achieve the same thing solely by applying type coercion using the custom conversion flag "`!j`" with the regular Python string formatting machinery, like
```ini
format = "{data!j}"
```
... which will serialize the value to JSON before putting it on the egress data path.

The change is not very intrusive, we included appropriate documentation for the README and hope you like this.

With kind regards,
Andreas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2392,6 +2392,31 @@ item = {
 
 ## Advanced features
 
+### JSON output serialization
+
+When receiving JSON data like `{"data": {"humidity": 62.18}}`, you might
+want to extract values using the `format` mechanism before forwarding
+it to other data sinks, like
+
+```ini
+format = "{data}"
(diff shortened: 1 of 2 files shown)


jpmens:
Merged, thank you!

PR #291, opened 2018-02-13
autoremote bridge service created

jusplainmike:
I wrote a service for autoremote that forwards mqtt messages to clients. 
https://joaoapps.com/autoremote/


jpmens:
Merged, thanks.

PR #304, opened 2018-04-07
Pushover images

psyciknz:
Add the image upload function as supported by pushover
- changed from urllib2 to requests for the loading of images
from a json payload via the "imageurl" attribute or by decoding a base64
encoded image in the the "image" attribute.
- text to accompany the
image comes from the "message" attribute of the json payload.


README.md
```diff
@@ -1885,6 +1885,26 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
 | `title`       |   O    | application title (dflt: pushover dflt) |
 | `priority`    |   O    | priority. (dflt: pushover setting)     |
 
+The pushover servive will accept either a simple mqtt message and post that to pushover.  Or a json pauload can be passed
```
sumnerboy12: Typo in `pauload` and I think it should be `including 1 *of* 2 image types`?

README.md
```diff
@@ -1885,6 +1885,26 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
...
+
+or
+```json
+ {
+    "message": "text message to post",
+    "imageurl": "full http url of an image to download and include"
+ }
+```
+
+There is an example of a format method that will introduce an image url from a particular mqtt topic.
```
sumnerboy12: Maybe call it a `format *function*`, and add the name of the sample function.

samplefuncs.py
```diff
@@ -71,6 +71,47 @@ def TopicTargetList(topic=None, data=None, srv=None):
 
     return targets
 
+def PushoverAddImageUrl(data=None, srv=None):
```
sumnerboy12: I think this should be removed actually. Pretty specific to your use-case and doesn't really fit the bill as a simple sample function. I think the examples you added in README.md are sufficient.

services/pushover.py
```diff
@@ -14,19 +22,29 @@
...
+        params['token'] = os.environ['PUSHOVER_TOKEN']
+    else:
+        params['token'] = token
+    #params['token'] = token
 
-    if not 'token' in kwargs:
-        kwargs['token'] = os.environ['PUSHOVER_TOKEN']
-    if not 'user' in kwargs:
-        kwargs['user'] = os.environ['PUSHOVER_USER']
+    #if user is None:
```
sumnerboy12: Why is this commented out? Should be included to keep same functionality as before?

services/pushover.py
```diff
@@ -14,19 +22,29 @@
...
 
-def pushover(**kwargs):
-    assert 'message' in kwargs
+def pushover(message, user, token, params, filepayload=None):
+
+    if token is None:
+        params['token'] = os.environ['PUSHOVER_TOKEN']
+    else:
+        params['token'] = token
+    #params['token'] = token
```
sumnerboy12: Remove commented code, we have git history for that ;).

services/pushover.py
```diff
@@ -14,19 +22,29 @@
...
+    #if user is None:
+    #    params=['user'] = os.environ['PUSHOVER_USER']
+    #else:
+    #    params['user'] = user
+    params["user"] = user
+
+    params["message"] = message
 
     url = urlparse.urljoin(PUSHOVER_API, "messages.json")
-    data = urllib.urlencode(kwargs)
```
sumnerboy12: Don't we still need to URL encode the request parameters?

services/pushover.py
```diff
@@ -14,19 +22,29 @@
...
-    req = urllib2.Request(url, data)
-    response = urllib2.urlopen(req, timeout=3)
-    output = response.read()
+
+    if filepayload is None:
+        r = requests.post(url,data=params, headers={'User-Agent': 'Python'})
```
sumnerboy12: Might be better to use `User-Agent: mqttwarn`?

services/pushover.py
```diff
@@ -77,9 +95,37 @@ def plugin(srv, item):
...
+            srv.logging.debug("Image is in json object %s" % messagejson["image"])
+            filepayload = { "attachment": ("image.jpg",base64.decodestring(messagejson["image"]), "image/jpeg") }
+            message = messagejson["message"]
+        else:
+            srv.logging.debug("No image could be found")
+
+
+
+    except Exception, e:
+        srv.logging.warn("Error sending pushover notification to %s [%s]: %s" % (item.target, params, str(e)))
```
sumnerboy12: This logging message is wrong - an exception here just means we didn't have a suitable JSON payload. Would be better to catch the decoding exception from `json.loads()` and `pass`, and handle any other exception with a suitable error message and fail the operation.

sumnerboy12: Think the expected error is `JSONDecodeError`

services/pushover.py
```diff
@@ -77,9 +95,37 @@ def plugin(srv, item):
     if callback is not None:
         params['callback'] = callback
 
+    srv.logging.debug("=================Starting Image processing")
```
sumnerboy12: Think it is worth reducing the amount of debug logging you have in here. Probably a bit much now you have completed development and proved it works.

services/pushover.py
```diff
@@ -14,19 +22,29 @@
 
 class PushoverError(Exception): pass
 
-def pushover(**kwargs):
```
sumnerboy12: Any reason you changed this? I am no python expert but I do like `**kwargs` - nice and concise and easy to extend (as you are doing here).

services/pushover.py
psyciknz: I think that's handled natively by requests in here:
```
if filepayload is None:
  r = requests.post(url,data=params, headers={'User-Agent': 'Python'})
else:
  r = requests.post(url,data=params, files=filepayload, headers={'User-Agent': 'Python'})
```

services/pushover.py
psyciknz: Yeah unsure.  The try will catch an error decoding the json, so yes the message to logging should change.   But as it is, it will also skip all the attempts at parsing that json that it couldn't read.

I think I'd be better to update the error to "could not parse json etc", and then fail out of the routine.

So like this:
```
except JSONDecodeError, jsone:
    srv.logging.error("Error decoding json message when processing images %s [%s]: %s" % (item.target, params, str(jsone)))
    return False
except Exception, e:
    srv.logging.error("Error parsing json parameters in sending pushover images to %s [%s]: %s" % (item.target, params, str(e)))
    return False
```

services/pushover.py
psyciknz: agrree

services/pushover.py
psyciknz: Yeah I'll have to think about this....probably could go back to how it worked now that its working....but when I started, this really threw me.  I'd not seen it before, maybe out the end of the method, but not the entire method definition.

services/pushover.py
psyciknz: So the only bit I'm unsure of is the final else:
```
else:
            srv.logging.debug("No image could be found,")
```
basically this will send out a json payload to pushover, as it's decoded it, but it didn't contain either of the two image formats.  I'm kind happy with that, should only happen if you screw up....unless I just pull the "message" attribute out and just send that as a text message only

services/pushover.py
```diff
@@ -77,9 +93,33 @@ def plugin(srv, item):
...
+            srv.logging.debug("Image url is in json object %s" % messagejson["imageurl"])
+            filepayload = { "attachment": ("image.jpg",requests.get(messagejson["imageurl"], stream=True).raw, "image/jpeg") }
+        elif 'image' in messagejson:
+            srv.logging.debug("Image is in json object %s" % messagejson["image"])
+            filepayload = { "attachment": ("image.jpg",base64.decodestring(messagejson["image"]), "image/jpeg") }
+        else:
+            srv.logging.debug("No image could be found,")
+			
+
+    except JSONDecodeError, jsone:
```
sumnerboy12: I think this should be a bit more forgiving of unexpected or invalid payloads. E.g. if you don't get a JSON payload then just pass the message as-is, this is how it currently works. If you get a JSON payload then check for the `message` attribute - if none found then pass the whole payload as-is. If there is a `message` attribute then check for your two image options, and if neither, or one of them fail to decode or resolve, then log a warning but still pass the message.

So we try and get a message delivered at all costs. Any issues get logged as warnings, but proceed. Sound reasonable?

services/pushover.py
sumnerboy12: TBH I had to look up the docs as I hadn't used it myself before. I am a bit of a Python noob as well. But it does seem like a good way to handle these types of functions with plenty of options. Up to you.

README.md
```diff
@@ -1885,6 +1885,25 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
 | `title`       |   O    | application title (dflt: pushover dflt) |
 | `priority`    |   O    | priority. (dflt: pushover setting)     |
 
+The pushover service will accept either a simple mqtt message and post that to pushover, or a json payload can be passed
+to the service including 1 of 2 image types and an accompanying message.
```
jpmens: What does "1 of 2 image types" mean? Is that a typo?

README.md
```diff
@@ -1885,6 +1885,25 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
...
+
+The following payload is for a base 64 encoded image.
```
jpmens: it's called "base64"; no space please

README.md
```diff
@@ -1885,6 +1885,25 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
...
+```json
+ {
+    "message": "text message to post",
+    "image": "base64 encoded image file"
+ }
+```
+
+or the following payload will download an image from an http service.  This must be accessible to the mqttwarn service.
```
jpmens: s/http/HTTP/g

samplefuncs.py
```diff
@@ -70,7 +70,7 @@ def TopicTargetList(topic=None, data=None, srv=None):
             targets.append('log:warn')
 
     return targets
-
+
```
jpmens: In future, please use one PR to solve one thing. Samplefuncs is not part of your PR so don't touch it. It's ok now, don't change this, but kindly keep in mind for any PR

services/pushover.py
jpmens: Definitely

README.md
psyciknz: there's two ways you can post an image.  Should I just replace that sentence with "can be passed to the service including a message and an image call"   and leave it up to the examples shown below to show the two difference types?

samplefuncs.py
psyciknz: Yeah I had altered it, to show how I incuded an image call from a regular message, but @sumnerboy12 suggested I leave it out as my examples were....so this is why samplefuncs has been touched in the pr (first I changed, then I removed my changes).

services/pushover.py
psyciknz: As mentioned, this broke my head when I tried switching back.  I'd probably need some guidance from you or @jpmens in how to reimplement it, if it's a better way of doing it.

PR #310, opened 2018-04-15
New service to insert SonOff messages into MySQL

halacs:
My new service (called halsql) inserts new record into MySQL.
This is a generic service, however, I designed it to colelct telemetry data from my SonOff POW devices.
This service can add static values (like the source/meaning of the data; e.g. 'bojler_enabled') and can remap keys (e.g. current power consumption data comes as 'current' but stored in DB as 'value').

Example MQTT message:

```
17:08:45 MQT: tele/bojler/STATE = {"Time":"2018-04-15T17:08:45","Uptime":"1T03:20:10","Vcc":3.221,"POWER":"ON","Wifi":{"AP":1,"SSId":"myHomeNetwork","RSSI":66,"APMac":"xx:xx:xx:xx:xx:xx"}}
17:08:45 MQT: tele/bojler/SENSOR = {"Time":"2018-04-15T17:08:45","ENERGY":{"Total":320.144,"Yesterday":5.105,"Today":1.881,"Period":0,"Power":2,"Factor":0.07,"Voltage":234,"Current":0.128}}
```

Example MySQL records:

```
mysql> select * from test;
+------------+-----------+----------------+
| timestamp  | value     | description    |
+------------+-----------+----------------+
| 1523804801 |   1.00000 | bojler_enabled |
| 1523804801 |   0.00000 | bojler_power   |
| 1523804801 |   0.00000 | bojler_current |
| 1523804801 | 235.00000 | bojler_voltage |
| 1523804801 |   0.00000 | bojler_factor  |
| 1523804801 | 320.14400 | bojler_total   |
+------------+-----------+----------------+
6 rows in set (0.00 sec)
```


jpmens:
Thank you for your contribution, but there are some issues I have to address:

1. The PR contains a file `services/*.sample` which doesn't belong there. The `services/` directory is for service modules.
2. Your PR has no addition to the README with its documentation. As such I will not take it. We take modules only when they've been documented so that others (including myself btw) know how to use them and why.
3. You have a `funcs.py` file in the PR, and it's not clear to me what that's for.

I think that last point in particular is going to be difficult to bring across, but I encourage you to give it a try.

halacs:
You are right! Thanks for the comments! Give me some hours (in worst case some days) and I will fix it.

1. I remove that file and move its content into the README.
2. I will do it. Sorry.
3. I think this also should be moved into the README instead. That func.py exists because I can visualize data with Grafana if timestamp is in unix format and "ON"/"OFF" text replaced with numbers (1/0). This is very SonOff specific.

jpmens:
You might want to place the file from 3. into `contrib/` into a directory which you prepare a working example, say.

halacs:
Hi,

I'm done.
I hope it looks fine now, however, if you have any comment please let me know.

Thanks & br,
Halacs

jpmens:
Merged, thank you

PR #314, opened 2018-07-19
generate a TOC and re-arrange the sections

rgitzel:
I find the README difficult to navigate, given its length. Yesterday while sorting out how to process JSON, I found myself wishing for an easier way to jump around (I had to use browser searching a lot). So I've added a table of contents! 

I used [`markdown-toc`](https://github.com/jonschlinkert/markdown-toc) like this:
```
markdown-toc README.md | pbcopy
```
Paste that into `README.md`, delete the first generated line, and remove the list of links to the plugins (it's redundant, there).

Then I saw why some of the reasons I was jumping around so much, so I re-ordered the sections.

Other than minor tweaks to some section titles, the only content change was to combine the diagram and text from the start of the plugins section into the introduction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,298 @@ To _warn_, _alert_, or _notify_.
 
 This program subscribes to any number of MQTT topics (which may include wildcards) and publishes received payloads to one or more notification services, including support for notifying more than one distinct service for the same message.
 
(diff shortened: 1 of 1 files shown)


jpmens:
This looks very good. Thank you!

jpmens:
Merged with thanks.

Issue #275, opened 2017-11-30
transform using a nested json value


kkplein:
This question is very basic, but part of my payload looks like this:

>     "inverterType": 3, 
>     "isDTSeries": false, 
>     "isOnline": true, 
>     "lastSeen": 1512028792882, 
>     "runningInfo": {
>         "eDay": 0.0, 
>         "eTotal": 10970.1, 
>         "errorMessage": [], 
>         "fac1": 50.03, 
>         "fac2": 50.02, 
>         "fac3": 50.02, 
>         "function": 129, 
>         "gcfiFault": 32, 
>         "hTotal": 4953, 
>         "iac1": 0.6, 
>         "iac2": 0.9, 
>         "iac3": 1.1, 
>         "ipv1": 0.6, 
>         "ipv2": 0.5, 
>         "line1FFault": 62.85, 
>         "line1VFault": 0.0, 
>         "line2FFault": 31.22, 
>         "line2VFault": 0.0, 
>         "line3FFault": 0.0, 
>         "line3VFault": 0.0, 
>         "pac": 80, 
>         "pv1Fault": 0.0, 
>         "pv2Fault": 0.0, 
>         "temp": 28.2, 
>         "tempFault": 121.2, 
>         "timestamp": "20171130075952", 
>         "vac1": 234.1, 
>         "vac2": 237.4, 
>         "vac3": 235.5, 
>         "vpv1": 275.1, 
>         "vpv2": 544.3, 
>         "workMode": 1
>     }, 
> 

...

I can use the 'first level' values in a transform (lastSeen, isOnline, etc) but I also would like to use the 'subvalues' values values from runningInfo, and I am not sure how to 'address' those... Tried runningInfo/vac2 but (of course...) that doesn't work.
Could anyone tell me how to use the value of vac2 in a transform?
Thanks!

kkplein:
After doing more research I know now that what I'm looking for is to access a nested json array. In python like this:
`print j['runningInfo'][1]['vac1']`
Question remains: how can I use this in a mqttwarn transform operation?

jpmens:
I think you'll have to use a custom Python function for that. See the README for how that works.

kkplein:
Hi Jan Piet,
bedankt voor je antwoord.

rgitzel:
In case anyone else finds this and wonders how the "custom Python function" works, I think I have it figured out.  The key is to provide an `alldata` function, which seems to be the only way to _change_ a payload.

[...]

PR #316, opened 2018-07-19
add an example of accessing deeper JSON nodes

rgitzel:
This is a simplified version of https://github.com/jpmens/mqttwarn/issues/275#issuecomment-406310145


jpmens:
I have merged this, @rgitzel, and thank you for it -- that's very good, and we greatly appreciate documentation contributions.

That said, this will probably change in the future, as we have been toying with the idea of implementing JSONpath which should make this much easier.

PR #317, opened 2018-07-24
clarify and simplify the example

rgitzel:
I realized over the weekend that I hadn't really stated the key to this, which is to create a new JSON, flat JSON document.

I've also simplified the example function.


jpmens:
Merged, thank you.

PR #319, opened 2018-07-26
device feature: notify only specific devices

r41d:
For now, support has only been added for pushover.

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,6 @@ I've written an introductory post, explaining [what mqttwarn can be used for](ht
   | `datamap`     |   O    | function name parse topic name to dict         |
   | `alldata`     |   O    | function to merge topic, and payload with more |
   | `format`      |   O    | function or string format for output           |
-  | `device`      |   O    | used by certain targets (see below). Usually a string |
   | `priority`    |   O    | used by certain targets (see below). May be func()  |
   | `title`       |   O    | used by certain targets (see below). May be func()  |
   | `image`       |   O    | used by certain targets (see below). May be func()  |
@@ -1985,7 +1984,7 @@ device = cellphone1,cellphone2
 targets = {
     'nagios'     : ['userkey1', 'appkey1', 'sound1'],
     'alerts'     : ['userkey2', 'appkey2'],
-    'tracking'   : ['userkey1', 'appkey2', 'sound3'],
+    'tracking'   : ['userkey1', 'appkey2', None, 'cellphone1,cellphone2'],
     'extraphone' : ['userkey2', 'appkey3']
     }
 ```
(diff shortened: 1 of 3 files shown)


r41d:
Not related to this PR but something I noticed: The first line of the main script should probably say
```
#!/usr/bin/env python2
```
with a 2 at the end, because the code is python2 only.

jpmens:
Thanks for your idea / contribution. Why does this addition require modifying mqttwarn.py and not simply use the pushover target configuration? In particular because it is used by pushover only.

r41d:
Good question. My first approach was to incorporate this into Pushover by changing the data structure from a list to a dict. The dict could then include 'sound' and 'device' if desired. However, this would break all existing configurations using Pushover, which is bad, so I started over.
Just like specifying a priority, I would consider it a generic feature, e.g. Pushbullet supports it too. I just didn't have experience with any other service, that's why I only included support for Pushover at this time.

jpmens:
Without having thought about this very much, I think this is too invasive as it stands, and I don't understand why you don't add this _device_ as an extra addr. Looking at the code of the current `pushover.py` service

```python
    # addrs is an array with two or three elements:
    # 0 is the user key
    # 1 is the app key
    # 2 (if present) is the PushOver sound to play for the message

    try:
        userkey = addrs[0]
        appkey  = addrs[1]
    except:
        srv.logging.warn("No pushover userkey/appkey configured for target `%s'" % (item.target))
        return False

...
    if len(addrs) > 2:
        params['sound'] = addrs[2]
```

it seems to me you could add it `if len(addrs) > 3` or similar.

jpmens:
Let people specify sound as `nil` or something, check and ignore that value.

r41d:
OK, I changed it, looking forward for review

README.md
```diff
@@ -1994,6 +1995,9 @@ notify, say, one or more of your devices as well as one for your spouse. As you
 can see in the example, you can even specify an optional sound to be played for
 the individual users. For a list of available sounds see the [Pushover API List](https://pushover.net/api#sounds).
 
+You can also specify the devices that should be notified, this is a comma-separated list.
```
jpmens: please add "comma-separated list of device names specified as a single string" to ensure people don't try to add a Python-like array or something.

mqttwarn.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
jpmens: Kindly undo that change for the time being. That path may be correct for your system but it isn't in general.

README.md
r41d: Sure, more clarification is always good

mqttwarn.py
r41d: Thought this was pretty general for Linux, but sure, will change it

jpmens: It may well be, but there's not just Linux in the world. :-)

jpmens:
Merged, thank you!

Issue #318, opened 2018-07-25
an official Docker image?


rgitzel:
I've been running `mqttwarn` in Docker on my laptop to experiment, and now plan to run it for real in Docker on my Synology NAS.  (I've got all my automation tools running in Docker: Mosquitto, Home Assistant, Node-Red; even InfluxDB and Grafana run just fine on the NAS.)  It's worked quite well, however, I've had to create my own Docker image. 

So I propose a `Dockerfile` be added to this project, and an image be published to Dockerhub whenever a new version is released.

There seems to be interest in such a thing, as there are 11 images [already in Dockerhub](https://hub.docker.com/search/?isAutomated=0&isOfficial=0&page=1&pullCount=0&q=mqttwarn&starCount=0), unfortunately most of them were one-offs and/or are quite old.

Fortunately, a good `Dockerfile` does already exist:  https://github.com/pfichtner/docker-mqttwarn  He just doesn't publish the images.

If @pfichtner agrees, his file could be added here. The only changes I would suggest is (1) to build the image from a GitHub tag, rather than from `master`, and (2) to include all the optional Python packges (like `jinga2`).  I can change the `Dockerfile` to take the tag as a parameter, making it easier to publish when `mqttwarn` is versioned.

pfichtner:
Please feel free to use and change my Dockerfile as needed

pfichtner:
Perhaps beside the tagged versions (foobar:version) you could also push master builds (foobar:latest) to docker hub?

amotl:
We will also provide pip-installable Python sdist packages on PyPI, see #127. We are almost there, i will try to push things further.

Just FYI: This might have an impact on how Docker images are built, specifically because we will employ solid versioning and repository tagging after this has been merged.

rgitzel:
@pfichtner I wasn't thinking straight:  now that the Dockerfile is in the same repo as the application, the Github tag doesn't matter.  The Docker image will always be built from master, and then tagged with the version (including `latest`) when it's deployed to Dockerhub

rgitzel:
I've created a PR: https://github.com/jpmens/mqttwarn/pull/320

For now, this is the @pfichtner Dockerfile, with a change to use the files currently checked out, rather than from Github directly.

This will work for the current version, and the Docker image can be published to Dockerhub for the existing release (0.10.1).

I will also work on a version of it based on the new `pip` work in `develop`.

rgitzel:
[...]

PR #320, opened 2018-07-27
add a Dockerfile

rgitzel:
Please see https://github.com/jpmens/mqttwarn/issues/318

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM python:2.7
+
+# based on https://github.com/pfichtner/docker-mqttwarn
+
+# install python libraries (TODO: any others?)
+RUN pip install paho-mqtt requests jinja2
+
+# build /opt/mqttwarn
+RUN mkdir -p /opt/mqttwarn
+WORKDIR /opt/mqttwarn
+
+# add user mqttwarn to image
+RUN groupadd -r mqttwarn && useradd -r -g mqttwarn mqttwarn
+RUN chown -R mqttwarn /opt/mqttwarn
+
+# process run as mqttwarn user
+USER mqttwarn
+
+# conf file from host
+VOLUME ["/opt/mqttwarn/conf"]
+
+# set conf path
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM python:2.7
+
+# based on https://github.com/pfichtner/docker-mqttwarn
+
+# install python libraries (TODO: any others?)
```
rgitzel: Are there others?

README.md
```diff
@@ -16,10 +16,6 @@ I've written an introductory post, explaining [what mqttwarn can be used for](ht
     + [Requirements](#requirements)
     + [Installation](#installation)
     + [Configuration](#configuration)
-      - [The `[defaults]` section](#the-defaults-section)
```
rgitzel: I manually removed more sub-sections, they were just cluttering up the TOC

README.md
```diff
@@ -3080,6 +3079,172 @@ pinger = 10.5; now=true
...
+```
+
+You can then edit any files and rebuild the image as many times as you need. 
+You don't need to commit any changes.
+
+The name `mqttwarn-local` is not meaningful, other than
+making it obvious when you run it that you are using your
+own personal image.  It will be renamed when it is published.
+
+### Deploy the Image (Administrators Only)
```
rgitzel: This doesn't seem an entirely appropriate location. At some point we'll have a separate document for contributors rather than users?

jpmens:
Thank you very much, Rodney, for this contribution. In particular, thank you for the verbose, and clear documentation you've added!

I am merging this now; we might have to change bits and pieces later, but that's fine, and feel free to add stuff as you go along.

Dockerfile
jpmens: lots, depending on the services people want to use. Ignore for now, as this will be changing later as @amotl already indicated.

README.md
jpmens: Yes

PR #322, opened 2018-08-08
InfluxDB plugin enhancements

jwyse:
Enhanced the InfluxDB plugin:

- Support InfluxDB retention policies (optional, global, and can override per target)
- Support InfluxDB precision (optional, global, and can override per target)
- Support writing to multiple InfluxDB databases (optional, can override the default database per target)
- Support InfluxDB tags and fields using transformation format string (optional, can specify per topic)

Sample ini snippet illustrating the new features:

```ini
[config:influxdb]
host      = 'influxdbhost'
port      = 8086
username  = 'username'
password  = 'password'
database  = 'mqttwarn'

# Retention Policy: optional (default: autogen)
rp        = 'retentionpolicy'
# Precision: optional (default: ns)
precision = 's'    # { ns, u, ms, s, m, h }

targets = {
                       # measurement (use database, rp, and precision specified above)
    'temperature'   : [ 'temperature' ],
                       # measurement,    database,   rp,     precision
    'disk'          : [ 'disk',          'servers',  'rp',   'h' ]
                       # measurement,    database   (default rp & precision)
    'cpu'           : [ 'cpu',           'servers' ],
                       # use default rp, but override database & precision:
    'alpha'         : [ 'alpha',         'metrics',  '',    's' ]
    }
```
```ini
[environment/temperature/basement]
targets = influxdb:temperature
format = room=basement,entity=sensor2 temperature={payload}
```

This should be backward-compatible with existing configurations using the InfluxDB plugin: it should work if `rp` and/or `precision` are not specified, and if targets only contain the minimum required value (measurement name).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1076,7 +1076,7 @@ This service provides a way for forwarding data to the time series database [Inf
 
 You will need to install an instance of InfluxDB (v9+) and create a new user. Then create a new database and give your user write permissions to that database.
 
-You can then setup multiple *targets*, each of which is a different *measurement* in your InfluxDB database.
+You can then setup multiple *targets*, each of which is a different *measurement* in your InfluxDB database.  Individual targets can override the default measurement, retention policy, and/or precision.
 
 Each time a value is received for an InfluxDB target, the value is sent to the configured *measurement* with a *topic* tag matching the MQTT topic the data arrived on.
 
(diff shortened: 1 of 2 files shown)


jpmens:
Merged. Thank you very much.

PR #335, opened 2018-09-17
clarify mysql target documentation - DON'T MERGE, HAS QUESTIONS

rgitzel:
https://github.com/rgitzel/mqttwarn/tree/issue-334-clarify-mysql-target#mysql

### DIFF
--- a/README.md
+++ b/README.md
@@ -1421,7 +1421,7 @@ If you added columns 'topic', 'payload' and '_dtiso' to the databse, then that s
 |   90 | Jane Jolie | { "name" : "Jane Jolie", "id" : 90, "number" : 17 } | 2018-09-17T20:20:31.889002Z | my/2  |
 +------+------------+-----------------------------------------------------+-----------------------------+-------+
 ```
-Here, the plugin pulled values for the new columns from standard mqttwarn meta-data.
+Here, the plugin pulled values for the new columns from standard mqttwarn metadata.
 
 When a message is received, the  plugin will attempt to populate the following column names:
 - root-level JSON nodes in the message
@@ -1444,8 +1444,8 @@ The MySQL plugin is one of the most complicated to set up.
 First it requires the [MySQLDb](http://mysql-python.sourceforge.net/) library to be installed, which is not trivial.
 - _Ubuntu 16.04:_
 ```
-    sudo apt-get install -y python-dev libmysqlclient-dev
-    sudo pip install MySQL-python
+sudo apt-get install -y python-dev libmysqlclient-dev
+®sudo pip install MySQL-python
 ```
 
 It then requires the following configuration section:
--- a/README.md
+++ b/README.md
@@ -1441,9 +1441,10 @@ To be clear, there is no other way to configure this particular plugin to use di
 
 The MySQL plugin is one of the most complicated to set up.
 
-First it requires the [MySQLDb](http://mysql-python.sourceforge.net/) library to be installed.
-- _Debian/Ubuntu_: `sudo apt-get install -y python-mysqldb`
-- TODO: do we have any others?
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1396,7 +1396,56 @@ string isn't available in the _data_, the message is _not_ published.
...
+Here, the plugin pulled values for the new columns from standard mqttwarn meta-data.
+
+When a message is received, the  plugin will attempt to populate the following column names:
+- root-level JSON nodes in the message
+  - e.g. 'name' and 'id' above
+- ['transformation data' fields](https://github.com/jpmens/mqttwarn#outbound-messages) names
+  - e.g. 'topic', 'payload' and '_dtiso' as above
+  - note that these all must be VARCHAR columns; timestamp columns are [not yet supported](https://github.com/jpmens/mqttwarn/issues/334#issuecomment-422141808)
+- the 'fallback' column, as noted below
+  - #TODO: how is this different from 'payload'?
```
rgitzel: @jpmens ?

README.md
```diff
@@ -1396,7 +1396,56 @@ string isn't available in the _data_, the message is _not_ published.
...
+ then you can use an `alldata` function to transform the incoming message into a JSON document with the
+ desired node names.
+
+#### Setup
+
+The MySQL plugin is one of the most complicated to set up.
+
+First it requires the [MySQLDb](http://mysql-python.sourceforge.net/) library to be installed.
+- _Debian/Ubuntu_: `sudo apt-get install -y python-mysqldb`
+- TODO: do we have any others?
```
rgitzel: @jpmens ?

This isn't the easiest library to install. :(

README.md
jpmens: the fallback column gets all the JSON elements which have **not** been otherwise stored. e.g. if `payload` contains `{ name: Jane, number: 42 }` (think JSON please), and we have a column `name` in the DB, then fallback will get `{number: 42}` whereas `payload` would have the full original payload.

rgitzel: I just tried it, and it _seems_ that 'payload' and 'full' are the same.  I added the 'full' column and posted a new message:
```
|   90 | Jane Jolie | { "name" : "Jane Jolie", "id" : 90, "number" : 17 } | NULL                                                | 2018-09-17T20:20:31.889002Z | db    |
|   91 | Jane Jolie | { "name" : "Jane Jolie", "id" : 91, "number" : 17 } | { "name" : "Jane Jolie", "id" : 91, "number" : 17 } | 2018-09-18T20:44:51.662836Z | db    |
```

Looking at the code, I see where the fallback column is set: https://github.com/jpmens/mqttwarn/blob/master/services/mysql.py#L67

But there's no point where fields are removed.  We only know which fields should be removed _after_ `add_row` is called, which is too late.

README.md
rgitzel: I found a way to do it with `pip` so changed the instructions.

README.md
```diff
@@ -1396,7 +1396,56 @@ string isn't available in the _data_, the message is _not_ published.
...
+If you added columns 'topic', 'payload' and '_dtiso' to the databse, then that same message will add this row:
+
+```mysql
++------+------------+-----------------------------------------------------+-----------------------------+-------+
+| id   | name       | payload                                             | _dtiso                      | topic |
++------+------------+-----------------------------------------------------+-----------------------------+-------+
+|   90 | Jane Jolie | { "name" : "Jane Jolie", "id" : 90, "number" : 17 } | 2018-09-17T20:20:31.889002Z | my/2  |
++------+------------+-----------------------------------------------------+-----------------------------+-------+
+```
+Here, the plugin pulled values for the new columns from standard mqttwarn meta-data.
```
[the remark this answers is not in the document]
rgitzel: Done.

jpmens:
Thank you very much. Merged.

PR #339, opened 2018-10-20
copy sample ini file to conf directory

tomhoover:
I am running mqttwarn in a Docker on [unRAID](https://unraid.net/), and using the `MQTTWARNINI` environment variable to change my configuration directory to a directory on the local host. This PR causes the `mqttwarn.ini.sample` file to be copied to the configuration directory on startup, if `mqttwarn.ini` doesn't exist.

### DIFF
--- a/mqttwarn.py
+++ b/mqttwarn.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import shutil
 import paho.mqtt.client as paho   # pip install paho-mqtt
 import imp
 import logging
@@ -337,6 +338,8 @@ try:
     cf = Config(CONFIGFILE)
 except Exception, e:
     print "Cannot open configuration at %s: %s" % (CONFIGFILE, str(e))
+    shutil.copyfile('mqttwarn.ini.sample', CONFIGFILE + '.sample')
+    print "\nA sample configuration file has been created at %s. Please rename to mqttwarn.ini and edit as appropriate." % (CONFIGFILE + '.sample')
     sys.exit(2)
 
 LOGLEVEL  = cf.loglevelnumber


jpmens:
Thank you very much for your contribution. I _think_ I know what you're attempting to do, but I'm very hesitant to take this. What I see is that you are creating a copy of the original exacmple file.

Why?

The reason for my hesitation is my dislike of software which does something that doesn't make sense to me. Try and "sell" me your idea please, in particular try to explain why you don't simply copy the file before starting.

tomhoover:
I appreciate your question--let me see if I can explain it better. As is often the case with Docker containers (and almost always when running containers on an unRAID system), the configuration directory is mapped to the underlying host file system. This patch causes the "sample" config file to be automatically copied from within the Docker container to the directory on the host system, when the user first attempts to "run" the container (making it easy for the user to create their own mqttwarn.ini from the sample). It just makes it a bit easier to set things up (since one cannot copy the sample file from "within" the container unless it's running).

What happens is that during the first "run" of the container, mqttwarn will not start due to the missing .ini file; however, it does copy the sample for the user, and informs the users to create their mqttwarn.ini from the sample. Once the do that, future "runs" will work as designed.

I hope that made better sense--if not, please let me know. Thanks!

amotl:
Dear Tom,

while your requirements resonate with me, there are a couple of things which come to my mind when thinking about this.
- It won't help users to reason about the behavior of the software, it's just not deterministic.
- Putting configuration files into the right place belongs to packaging, deployment, installation or configuration. It shouldn't be part of the software itself.
- I don't know of any software which copies and activates configuration files on its own behalf.
- Activating a default configuration file could even be dangerous, depending on its contents.
- The system the software is running on might not allow any write access.

I'm sure others can explain my concerns more concise.

However, i believe your problem can easily be solved when looking at it from a different perspective.

Cheers,
Andreas.

jpmens:
@tomhoover thank you for elaborating, and I understand your reasoning, but we won't take this.

I dislike the idea of a program copying a configuration file and activating that for me, but I explicitly thank you for the contribution idea. Keep them coming!

amotl:
Dear Tom,

i am beginning to see your point. However, i am holding my argument that this procedure could obfuscate the way things are turning out. In any case, the user would have to be notified somehow what is going on and what would be the preferred steps in making _mqttwarn_ do what she means. This is also called "documentation" ;].

The polished mqttwarn release being in the making on behalf of the "develop" branch already has a command for creating a configuration file, just try `mqttwarn make-config` (its output can easily be redirected into a file). Maybe we should add an appropriate notice to the error message _mqttwarn_ would croak with when being launched without any configuration?

Would this make sense to you, @tomhoover and @jpmens? Thanks for bearing with us!

With kind regards,
Andreas.

tomhoover:
Hi Andreas, 

As I am attempting to move almost all utilities like this to docker containers, the purpose of my PR is to make deployment of a docker container as simple as possible. If one is running mqttwarn "normally" (i.e. outside of a docker container), I can see `mqttwarn make-config` as a very straightforward way to create a configuration file. One of the beauties of the docker method is that simply typing `docker run ...` downloads the image, and starts the container, automagically. 😄 

If one is creating a docker container from a docker image, here are the steps that would be followed using "my way" and "your way":

### My way:

`docker run --name=mqttwarn -e MQTTWARNINI=/opt/mqttwarn/conf/mqttwarn.ini -v ~/tmp/mqttwarn_conf:/opt/mqttwarn/conf tomhoover/mqttwarn`

which offers the following error message:

_Cannot open configuration at /opt/mqttwarn/conf/mqttwarn.ini: [Errno 2] No such file or directory: '/opt/mqttwarn/conf/mqttwarn.ini'_

_A **sample** configuration file has been created at /opt/mqttwarn/conf/**mqttwarn.ini.sample**. **Please rename to mqttwarn.ini and edit as appropriate.**_

One then simply types:

`mv mqttwarn.ini.sample mqttwarn.ini && vi mqttwarn.ini && docker start mqttwarn`

And they're up and running.

### Your way:

`docker run --name=mqttwarn -e MQTTWARNINI=/opt/mqttwarn/conf/mqttwarn.ini -v ~/tmp/mqttwarn_conf:/opt/mqttwarn/conf tomhoover/mqttwarn`

which offers the following error message:

_Cannot open configuration at /opt/mqttwarn/conf/mqttwarn.ini: [Errno 2] No such file or directory: '/opt/mqttwarn/conf/mqttwarn.ini'_

`docker rm mqttwarn`

`docker run -it --name=mqttwarn -e MQTTWARNINI=/opt/mqttwarn/conf/mqttwarn.ini -v ~/tmp/mqttwarn_conf:/opt/mqttwarn/conf tomhoover/mqttwarn python mqttwarn make-config > /opt/mqttwarn/conf/mqttwarn.ini`

`vi mqttwarn.ini && docker start mqttwarn`

**Summary:**

Both ways work; however, the first way takes considerably less typing. :-)

No worries--I'm content that we agree to disagree. I haven't yet had a chance to take a look at the develop branch, but I look forward to reviewing it and maybe making a contribution to it.

Thanks!

rgitzel:
I run under Docker as well. (And wrote the 'official' Dockerfile.)  

My concern with "my way" is that I don't like applications which create files without my asking.  In this case, if I don't run the program correctly (which often happens when I'm new to an application) my system will start to be littered with `mqttwarn.ini` files that I have to go clean up after.  :-(

Worse, if file permissions aren't right, not only will I get a message about a sample file being created for me, I will also get errors about not being able to create a file, which will both be confusing (WHY is it writing files?) and alarming (WHAT is it writing on my system?). 

"Your way" does seem more verbose, but it's also explicit.  And it's still not the application creating the file, it's the result of piping the output.  So both generating the content and the writing of the file are under my control.

It's also the way I've seen done with apps like InfluxDb and Telegraf.

rgitzel:
The "Your way" method might actually be _shorter_  as `mqttwarn` doesn't need a config folder to `make-config`:
```
docker run -t tomhoover/mqttwarn python mqttwarn make-config > /opt/mqttwarn/conf/mqttwarn.ini
```

Also, given you're running `mqttwarn` in a Docker container, is there actually a need to be explicit about the configuration folder? _Within_ the container, nothing else will interfere or care about the `/opt/mqttwarn/conf/` folder.

It's on the host that you typically want to put the config somewhere else (mine is at `/volume/docker/mqttwarn`), and that's controlled with the `-v` argument.

tomhoover:
docker run -t tomhoover/mqttwarn python mqttwarn make-config > /opt/mqttwarn/conf/mqttwarn.ini

**I like it.** 😄 

Please note, in my example above, I was also putting the config on the host by using:

`-v ~/tmp/mqttwarn_conf:/opt/mqttwarn/conf` (although in "real life", it resides at /mnt/cache/appdata/mqttwarn on my unRAID server)

PR #340, opened 2018-10-20
use MQTTWARNLOG, if set

tomhoover:
mqttwarn ignores the `MQTTWARNLOG` environment variable, if set, and instead uses the logfile location set within the configuration file.

This PR determines the location of the logfile by using the following order of precedence:

1. `MQTTWARNLOG` environment variable
2. logfile as defined in the configuration file

### DIFF
--- a/mqttwarn.py
+++ b/mqttwarn.py
@@ -340,7 +340,10 @@ except Exception, e:
     sys.exit(2)
 
 LOGLEVEL  = cf.loglevelnumber
-LOGFILE   = cf.logfile
+
+# if MQTTWARNLOG env variable exists, use it; otherwise, use config file
+LOGFILE    = os.getenv(SCRIPTNAME.upper() + 'LOG', cf.logfile)
+
 LOGFORMAT = cf.logformat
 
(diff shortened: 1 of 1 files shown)


mqttwarn.py
```diff
@@ -340,7 +340,10 @@ def join(self):
     sys.exit(2)
 
 LOGLEVEL  = cf.loglevelnumber
-LOGFILE   = cf.logfile
+
+# if MQTTWARNLOG env variable exists, use it; otherwise, use config file
```
jpmens: Would you please the word "config" to "configured"? I think that's what you mean.

tomhoover: Does this sound better? (I think it better explains it) If so, let me know and I'll push the change. Or, I can just change 'config' to 'configured', if you prefer that wording.

`# if MQTTWARNLOG env variable exists, use it; otherwise, use 'logfile' option from mqttwarn.ini`

jpmens: That's good.

jpmens:
Merged, thank you!

PR #345, opened 2018-11-11
Pushover images

psyciknz:
I'm made some modifications to the pushover service where the camera is secured by auth or digest authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1888,7 +1888,7 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
 The pushover service will accept a payload with either a simple text message, or a json payload which contains
 a `message` and either an `imageurl` or `imagebase64` encoded image.
 
-Further, the imageurl payload, can have the addition parameters of an auth type (basic, digest) and a user and password.
+Further, the imageurl payload, can have the addition parameters of an auth type (basic, digest) and a user and password.  This is useful if your imaging device uses authentication.  Some examples are some IP cameras, or some other simple internet based image services.
 
 The following payloads are valid;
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1914,6 +1916,17 @@ Simple text message
...
+```json
+ {
+    "message": "Message with image downloaded from URL: digest authentication",
+    "imageurl": "<image url>",
+    "auth": "digest",
+    "user": "myspecialuser",
+    "password": "myspecialpassword"
+ }
+```
+For the above example, I would only recommend this be used in a local MQTT server instance, as your password is being transmissed for your imaging device.
```
jpmens: `transmissed` is a typo. ;)

I think you'll want to add "transmitted in the clear" so that people understand.

README.md
```diff
@@ -1888,6 +1888,8 @@ NOTE: `callback` is an optional URL for pushover to [ack messages](https://pusho
 The pushover service will accept a payload with either a simple text message, or a json payload which contains
 a `message` and either an `imageurl` or `imagebase64` encoded image.
 
+Further, the imageurl payload, can have the addition parameters of an auth type (basic, digest) and a user and password.
+
```
jpmens: Please elaborate what this is actually used for. Why should I add this?

psyciknz: I've added some context as to why.

psyciknz: I've added some context to why this might be useful.

Sorry I can't figure out how to add the relevant code section that was updated in the last commit.

README.md
psyciknz: And this I've made clearer also. 

Again sorry, I don't know how to add the relevant section

jpmens:
Merged, thank you.

PR #351, opened 2018-11-29
Slack images

psyciknz:
I've added slack image support.  Formats are the same as pushover and shown in the readme.md

It does require slacker 0.10.0 or above.

I tried updating my docker file to get 0.10.0 rather that the current 0.95 that is installed but it had issues on arm and arm64.  

Also there are some extra permissions required for uploading images etc.  This isn't covered in the readme as yet I notice.


jpmens:
Looks good to me. Thank you.

Issue #358, opened 2019-01-18
function string type with mqttwarn in container


przemas75:
Hello,
Im a user of owntracks and tried to move mqttwarn to container.
However 'datamap' with a function OwnTracksTopic2Data() (i have used default one) does not work, it throws me  a warning:

`WARNING [mqttwarn] Cannot format message: u'username' `

This goes also for deviceid from topic.
I have found in docker container the topic is not type string but unicode.

rgitzel:
Hi @przemas75, 

"move to container" means running this in Docker? https://hub.docker.com/r/jpmens/mqttwarn

What version of `mqttwarn` were you using before?  It's odd the validation logic would have changed.

Thanks.

przemas75:
Yes, i meant running this in Docker (actually i build this myself, but also tested official container).
And unfortunately, my previous mqttwarn was VEEERY old... probably 3 years... 

After seeing this not working (in docker container) i run it from tar.gz, then checked official container. For me it was not working. 
But this function did not change since these years.

jpmens:
@przemas75 Sorry, but I am a bit confused. Does this or does this not work when you run it outside of Docker? Which version of Python?

przemas75:
It does work without container. In a container it showed the errors i posted (for function OwnTracksTopic2Data).
Default python in Debian 2.7.13 and in container i believe it was 2.7.15.

doebi:
I am having the exact same issue, either with or without docker. Trying to extract information from the topic to use it as tags for influxdb.

Any help appreciated.

Python v2.7.16 outside of docker

rgitzel:
Christoph,

Is this showing up when you use the Dockerfile in the repo?  Or are you
making your own?

rodney

doebi:
Hi rodney, thanks for your reply.

I am using docker-compose with "jpmens/mqttwarn:latest" as image definition.
I ust checked, the latest image was built 9 months ago.

Is the docker image in public repos up to date?

doebi:
well yeah. I just removed the line with type check: https://github.com/jpmens/mqttwarn/blob/master/samplefuncs.py#L12

...and now it's working. Why did you check for simple strings in that example func anyway?

jpmens:
I don't know. Would you be so kind as to make a PR which changes that?

doebi:
Ok. I just thought, maybe there is some reason or deeper meaning to it.

PR #359, opened 2019-01-18
adding additional type for topic check

przemas75:
Issue https://github.com/jpmens/mqttwarn/issues/358

### DIFF
--- a/samplefuncs.py
+++ b/samplefuncs.py
@@ -9,6 +9,16 @@ except ImportError:
     import simplejson as json
 
 def OwnTracksTopic2Data(topic):
+    if type(topic) == unicode:
+        try:
+            topic = str(topic)
+            parts = topic.split('/')
+            username = parts[1]
+            deviceid = parts[2]
+        except:
+            deviceid = 'unknown'
+            username = 'unknown'
+        return dict(username=username, device=deviceid)
     if type(topic) == str:
         try:
             # owntracks/username/device
--- a/services/twitter.py
+++ b/services/twitter.py
@@ -22,12 +22,19 @@ def plugin(srv, item):
 
     text = item.message[0:138]
     try:
+        text = text.decode('utf-8')
+    except Exception,  err:
+        srv.logging.debug("Cannot convert to unicode: %s" % (str(err)))
+    try:
         srv.logging.debug("Sending tweet to %s..." % (item.target))
         res = twapi.PostUpdate(text, trim_user=False)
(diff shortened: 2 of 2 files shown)


services/twitter.py
```diff
@@ -21,13 +21,20 @@ def plugin(srv, item):
...
     try:
         srv.logging.debug("Sending tweet to %s..." % (item.target))
         res = twapi.PostUpdate(text, trim_user=False)
         srv.logging.debug("Successfully sent tweet")
     except twitter.TwitterError, e:
         srv.logging.error("TwitterError: %s" % (str(e)))
         return False
+    except UnicodeDecodeError:
+        srv.logging.error("Your message could not be encoded.")
+    return False
```
jpmens: are you quite sure this `return False` is correct? (indentation)

przemas75: Thanks for this catch.

samplefuncs.py
```diff
@@ -9,6 +9,16 @@
     import simplejson as json
 
 def OwnTracksTopic2Data(topic):
+    if type(topic) == unicode:
```
jpmens: why is this required, and can you not merge this if together with the one on line 22?

services/twitter.py
```diff
@@ -21,13 +21,20 @@ def plugin(srv, item):
     )
```
jpmens: it is unfortunate, that you are mixing two different things in one pull request; please try to avoid that in future. (owntracks in samplefuncs.py has nothing to do with twitter.py)

services/twitter.py
```diff
@@ -21,13 +21,20 @@ def plugin(srv, item):
...
 
     text = item.message[0:138]
+    try:
+        text = text.decode('utf-8')
+    except Exception,  err:
+        srv.logging.debug("Cannot convert to unicode: %s" % (str(err)))
```
jpmens: are you sure this message is correct? that you are converting TO unicode?

przemas75: I think i understand why you questioning that. I meant converting to unicode string from string which contains non-ascii, it would be 'changing' meaning rather
As i remember 'text' had issues with non-ascii chars as a string.

services/twitter.py
przemas75: Yes, i understand that.

samplefuncs.py
przemas75: Because in my situation, topic was not type of string but unicode. I have simply checked this.
Im afraid i dont understand this well.

PR #366, opened 2019-01-28
fix #356

### LINKED ISSUE #356: pushsafer not work
I get an "WARNING [pushsafer] Error sending pushsafer notification to ..." when try to use pushsafer service
I am new in python but I try to find why. 
So look at pushsafer.py and I saw  
def pushsafer(**kwargs):
    #assert 'message' in kwargs   ??? why  check 'message' code add 'm' key instead ??? 
     assert 'm' in kwargs   #this work

    #if not 'privatekey' in kwargs:  ??? why  check 'privatekey' code add 'k' key instead, also 'k' always has a value or ''  ??? 
       #kwargs['privatekey'] = os.environ['PUSHSAFER_TOKEN']
   #following work
    if not kwargs['k']:
        kwargs['k'] = os.environ['PUSHSAFER_TOKEN']

dka2:
As described in issue #356

### DIFF
--- a/services/pushsafer.py
+++ b/services/pushsafer.py
@@ -12,10 +12,10 @@ PUSHSAFER_API = "https://www.pushsafer.com/"
 class PushsaferError(Exception): pass
 
 def pushsafer(**kwargs):
-    assert 'message' in kwargs
+    assert 'm' in kwargs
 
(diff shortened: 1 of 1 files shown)


jpmens:
Merged, thank you.

Issue #344, opened 2018-10-30
Official docker image need more modules


tinuva:
I have been following issue [341](https://github.com/jpmens/mqttwarn/issues/341) and is really excited to be able to use it, however I think it needs some of the prerequisites added to the Dockerfile.

For example the following log entries work on my standalone install:
```
mqttwarn        | 2018-10-30 11:19:16,856 DEBUG [slack] *** MODULE=services/slack.py: service=slack, target=openhabalarm
mqttwarn        | 2018-10-30 11:19:16,857 ERROR [slack] slacker module missing
mqttwarn        | 2018-10-30 11:19:16,858 WARNING [mqttwarn] Notification of slack for '/openhab/notification/alarm' FAILED or TIMED OUT
mqttwarn        | 2018-10-30 11:19:16,859 DEBUG [mqttwarn] Job queue has 5 items to process
mqttwarn        | 2018-10-30 11:19:16,859 DEBUG [mqttwarn] Processor #0 is handling: 'pushbullet' for david
mqttwarn        | 2018-10-30 11:19:16,864 DEBUG [pushbullet] *** MODULE=services/pushbullet.py: service=pushbullet, target=david
mqttwarn        | 2018-10-30 11:19:16,865 WARNING [pushbullet] pushbullet is not installed
mqttwarn        | 2018-10-30 11:19:16,866 WARNING [mqttwarn] Notification of pushbullet for '/openhab/notification/alarm' FAILED or TIMED OUT
````

On the normal install I am able to install the dependencies, but the docker image mean I need to build my own for my specific needs unless you could add some of the popular ones at least.

Many thanks in advance!

jpmens:
We cannot and will not add all prerequisites for all 60+ modules to the Docker image, and "popular" would appear to be in the eyes of the beholder, no?

rgitzel:
@jpmens  Actually, why _not_ add them all?  At worst it makes the Docker build time slightly longer, and the image slightly bigger. Any that you don't use yourself just sit there unused. And they don't interfere with your local Python environment.

It might just mean a bit of playing whack-a-mole explicitly listing them all.

tinuva:
> We cannot and will not add all prerequisites for all 60+ modules to the Docker image, and "popular" would appear to be in the eyes of the beholder, no?

Which ones will work currently? 

I think email works, but if that is the only one, I am not quite sure what the use of this docker image will be.

Please consider adding all if not most.

amotl:
Hi there,

[...]

PR #370, opened 2019-06-27
Created a Cloudflare DNS updating service

psyciknz:
Using the publish_public_ip from samplefuncs.py which posts an IP address of your network, you can then update your cloudflare A record dns entries.  Kind of a replacement to DynDNS or Noip services.

I did originally have python 3.7 set, and it recommended that some of the MQTTwarn.py Exceptions change from:
```
Exception exception, ex:
```
to
```
Exception exception as ex:
```
So that is why that is altered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,7 @@ _mqttwarn_ supports a number of services (listed alphabetically below):
 * [autoremote](#autoremote)
 * [carbon](#carbon)
 * [celery](#celery)
+* [cloudflare](#cloudflare)
 * [dbus](#dbus)
 * [dnsupdate](#dnsupdate)
 * [emoncms](#emoncms)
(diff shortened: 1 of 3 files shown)


jpmens:
Thanks for your contribution. Before I review this, would you please explain what the use case is, in particular as you'll be using builtin `publish_public_ip_address()` to obtain the IP address which you then submit to Cloudflare? At least that's what I'm understanding.

Why would I need this mqttwarn service?

psyciknz:
Basically it's a poor mans dynamic DNS service.  Ie if you use noip and use their free account (which has an 30 day expiration), or dyndns, which is being shut down, this is a replacement for that.

It just seemed a logical place for me to do it, rather than writing something else controlled by openhab/homeassistant to drive the cloudflare api changes to assign an ip address to a dns entry.

rgitzel:
You could accomplish the same thing with a shell script and a cron job?  How does using MQTT help?

psyciknz:
> You could accomplish the same thing with a shell script and a cron job? How does using MQTT help?

It doesn't do it any better or worse....and yes you could do it with cron + shell.  But then you have configuration stuff all over the place.  Where this means most of my external connectivity to outside services are limited to mqttwarn.  Just makes your system (IMO) more self contained.

psyciknz:
So.....maybe this is a case where it needs to be a "personal" service, and we need to further the options for bundled services and ones that users can add, ie a separate directory that can load custom services?

https://github.com/jpmens/mqttwarn/issues/344

jpmens:
I think you're on the right track with us needing a method of providing for "personal" services. This is possible today of course, by dropping the service and its requirements into the `services/` directory.

As to this particular example I don't see widespread use for it. While you did trigger my interest by using the magic incantation (`"dns"` :-) ) it's not something I think we should provide in a default install.

psyciknz:
Ok I’ll close.   Regarding doing local do we want to discuss that?

I know with docker I can just expose the services folder and introduce whatever I want.

But I was wondering about loading modules from services.   Ie built in.  And another folder services.local where users can put their own service files.  I actually already have two, this cloudflare one and an dns gateway that I use.

Equally I can just clone the directory and introduce them before building the docker container.   But I prefer the two folders method as cleaner.

psyciknz:
Cancelled.

amotl:
Hi again,

thank you for this conversation and contribution. The code here has just been added to the [`mqttwarn-contrib`] project, see also https://github.com/jpmens/mqttwarn/issues/435#issuecomment-864401349.

With kind regards,
Andreas.

amotl:
Hi David,

your extension is now available within the most recent `mqttwarn-full` Docker image.
```shell
docker pull ghcr.io/jpmens/mqttwarn-full:0.26.0

docker run -it --rm \
    ghcr.io/jpmens/mqttwarn-full:0.26.0 \
    python -c "import mqttwarn_contrib.services.cloudflare_zone as plugin; print(plugin.__author__)"
```

With kind regards,
Andreas.

PR #373, opened 2019-08-02
remove type check

### LINKED ISSUE #358: function string type with mqttwarn in container

doebi:
Updated sample function. Closes #358

### DIFF
--- a/samplefuncs.py
+++ b/samplefuncs.py
@@ -9,17 +9,15 @@ except ImportError:
     import simplejson as json
 
 def OwnTracksTopic2Data(topic):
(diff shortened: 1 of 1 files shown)


jpmens:
Merged, thank you.

PR #375, opened 2019-09-09
Local services

psyciknz:
A WIP, but here's a concept for having local services.  I envisage the standard folder with all the services in the repo.  But the user can also specify a local directory and local services to launch.

For docker, what I'll need to work out, is how any extra libraries might need to be loaded (probably a requirements file....maybe a dynamically loaded requirements/pip install when running).

TODO: I think I need to update the readme and remove the cloudflare


psyciknz:
I think I've managed to figure out a way.  The problem with updating the docker file, or using a requirements file is they're added at docker build stage.  

So what I did was change the entry point from calling `python mqttwarn.py` to calling a run.sh script.
By default, it's included, and just runs the python command.

But if you override the run.sh in docker with a volume mount eg:
```
 -v /users/mqttwarn/folder/run.sh:/opt/mqttwarn/run.sh
```
Then the users local copy of run.sh is executed, where they can add a pip install <new package here> command.

The other override if the location of the local services:
```
-v /user/mqttwarn/folder/local-services:/opt/local-services
```
Where the user has `localservices = /opt/local-services` in their `mqttwarn.ini`

psyciknz:
I've just used this locally (via docker).  So think it's ready for review.

Let me know if I've added enough explanation about how to set it up.

rgitzel:
@psyciknz It would be helpful, both for testing and evaluating this PR, and for future users, if you could provide an example or three.  

The files could be put in `examples` and then you could use that folder as an example when creating the configuration.

psyciknz:
Ok I''ll see about an example plugin

amotl:
Dear @psyciknz,

apologies for being that late on your request, as I found your scenario very reasonable. #496 now improves the situation, you can find the corresponding documentation for loading external services at [1].

Please let us know what you think about it and if that satisfies your use case. If you see any further things to be improved, please open a separate issue.

With kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/blob/master/HANDBOOK.md#loading-external-services

psyciknz:
Hi, I'll have to spin up a new instance using the default code (rather than my own version).
Is this code merged with a particular branch/docker tag?  And then I'll see if the extra services I had will load

amotl:
Hi @psyciknz,

the latest code is part of the most recent release, version 0.21.0. However, I believe we did not build and release any Docker image for that. Maybe @jpmens can do it?

With kind regards,
Andreas.

jpmens:
Docker images are built automatically: [0.21.0 was pushed 13h ago](https://hub.docker.com/r/jpmens/mqttwarn/tags?page=1&ordering=last_updated)

amotl:
> Docker images are built automatically.

Indeed! Thanks.

```
docker pull jpmens/mqttwarn
docker run -it --rm jpmens/mqttwarn mqttwarn --version
mqttwarn 0.21.0
```

amotl:
Dear @psyciknz,

I just verified it with the most recent Docker release, by reusing configuration from the repository itself. So, in order to reproduce, invoke those commands from the repository root:
```shell
docker run -it --rm --volume=$PWD:/etc/mqttwarn --env=MQTTWARNINI=tests/selftest.ini jpmens/mqttwarn
```

Then, you will see an output like
```log
2021-06-03 12:42:26,055 DEBUG    [mqttwarn.core            ] Trying to load service "tests.acme.foobar" from module "tests.acme.foobar"
2021-06-03 12:42:26,063 INFO     [mqttwarn.core            ] Successfully loaded service "tests.acme.foobar" from module "tests.acme.foobar"
```
```log
2021-06-03 12:42:26,063 DEBUG    [mqttwarn.core            ] Trying to load service "tests/acme/foobar.py" from file "tests/acme/foobar.py"
2021-06-03 12:42:26,072 INFO     [mqttwarn.core            ] Successfully loaded service "tests/acme/foobar.py"```
```

This indicates both variants of importing (by module vs. by file) outlined at [1] seem to work properly.

With kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/blob/master/HANDBOOK.md#loading-external-services

psyciknz:
> Docker images are built automatically: [0.21.0 was pushed 13h ago](https://hub.docker.com/r/jpmens/mqttwarn/tags?page=1&ordering=last_updated)

Ahh right, like me you haven't found a way of automatically creating amd64, arm32 and arm64.  I did have one for a while but it stopped.  If it works again, I can give you the scripts etc (it uses travis CI to build)

amotl:
Hi @psyciknz,

since integrating #509, the machinery automatically builds and published Docker images for amd64, arm64 and arm32. Enjoy!

With kind regards,
Andreas.

Issue #294, opened 2018-03-13
Issues with Formatting json


the-butterfry:
I'm trying to get the Format to correctly parse out json received with the File handler plugin.
Here's an example string that is being received.
{"Time":"2018-03-13T20:16:55","DS18B20":{"Temperature":19.8},"TempUnit":"C"}

All I want to do is pull out the {Temperature} data into the File, discarding the rest. I can't seem to make it work. Any tips?
P.S.
I can make {time} work just fine, but as soon as I try to transform {temperature} is where the problems starts.

amotl:
Dear Cory,

thanks for reporting this. It is probably about the fact that "Temperature" is nested inside the main dictionary, right? Accessing the data through JSON Pointer might help here in the future, see also #303.

With kind regards,
Andreas.

HillNet:
Hi,

I've been using mqttwarn to parse mqtt topic's to influxdb. Works perfectly.

I'm now looking at parsing nested JSON messages and this is where I get stuck (I'm a self taught hobbiest).

I see there is a workaround for nested JSON included in README.md. Would you be able to advise if that will work with multiple payloads? This is my received JSON message:

`{
   "Time":"2018-08-14T07:21:45",
   "DS18B20-1":{
      "Id":"051693DD63FF",
      "Temperature":15.4
   },
   "DS18B20-2":{
      "Id":"0516941253FF",
      "Temperature":14.8
   },
   "SI7021":{
      "Temperature":12.2,
      "Humidity":99.9
   },
   "TempUnit":"C"
}`

I'm happy to "have a go" but if you could comment I'd appreciate it.

Many thanks,

Roland

rgitzel:
@HillNet I wrote the workaround in the README. It will work for your situation -- multiple temperature sensors with just one Sonoff?  nice, I want to try that one day :-) -- it will take just a little more coding.

The problem is that one message can create only one InfluxDb entry, by default. What you want is for each message to be processed twice. So we'll make mqttwarn do that.

Make two Python functions like in the example, one pulling out the "-1" value and the other pulling out the "-2" value.  Then create two topic sections in `mqttwarn.ini`, but don't name them based on the topic (see https://github.com/jpmens/mqttwarn#the-__topic__-sections).  "Temperature1" and "Temperature2" would do fine.

That should do it.

HillNet:
Thanks @rgitzel , I'll have a go!
[...]

PR #376, opened 2019-10-05
Build post data correctly when no additional influx tags included

jacques42:
Links to issue #294 Issues with Formatting json

Enhancing the logic to build the data payload for Influx post in a way that format strings can work without having to add additional tags.


jpmens:
Merged, thank you.

Issue #395, opened 2020-01-12
service mqttpub has encoding trouble


jpmens:
```
WARNING  [mqttwarn.services.mqttpub] Cannot PUBlish via `mqttpub:mout': 'bytes' object has no attribute 'encode'
```

amotl:
Thanks for fixing that one as a leftover to #336!

Issue #384, opened 2019-11-20
Properly support UTF-8 everywhere


amotl:
When working on one of the next iterations, we should check and adjust for proper UTF-8 support within mqttwarn core as well as relevant service plugins.

See also #154, #360 and maybe others.

Issue #360, opened 2019-01-18
Twitter service croaks with charset encoding issue


przemas75:
Twitter service while working in container (mqttwarn) does not work as message cannot be sent due to 'type' issues.
`2019-01-17 21:57:26,839 ERROR [twitter] Error sending tweet to sss: 'ascii' codec can't decode byte 0xc5 in position 28: ordinal not in range(128)`
It's fixed in this pull https://github.com/jpmens/mqttwarn/pull/359

amotl:
Dear @przemas75,

we recognize *mqttwarn* might not be fully UTF-8 safe yet. So, in order to assist investigating it, I am humbly asking you to submit a reproducible example using the new feature to invoke a service plugin directly from the command line as outlined through [1].

Thanks already and with kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/tree/develop#running-notification-plugins

amotl:
Dear @przemas75,

your issue might have been resolved through `mqttwarn 0.12.0`. You should be able to install this using `pip install mqttwarn` on Python2.

With kind regards,
Andreas.

amotl:
Dear @przemas75,

mqttwarn might even be compatible with Python 3 these days. May we humbly ask you to test this again with the Twitter service?

Thanks already and with kind regards,
Andreas.

jpmens:
Closing for lack of activity.

Issue #154, opened 2015-11-12
instapush: non ascii characters cause an exception Parameter status cannot be formatted


AlexBGoode:
Hello

I managed to setup subscribe to mosquitto and instapush works as needed until I start using international messages, i.e. messages with non ascii characters

In my case it is Russian and when I send a message:
   **mosquitto_pub -h 127.0.0.1 -t test -m '{"status":"Привет"}'**
then I get the following error:
   **DEBUG [instapush] Parameter status cannot be formatted: 'ascii' codec can't encode characters in position 0-5: ordinal not in range(128)**

Tried with German just for example:
   **mosquitto_pub -h 127.0.0.1 -t test -m '{"status":"Straße"}'**
then I get the following error
   **DEBUG [instapush] Parameter status cannot be formatted: 'ascii' codec can't encode character u'\xdf' in position 4: ordinal not in range(128)**

Any ideas what's wrong with that? I believe most of us have to use non typical ascii characters

jpmens:
I'm ashamed to say, I don't think I've tested most of the modules with UTF-8. I _assume_ (and hope) the problem is within the instapush module.  I cannot verify this now, but maybe you can have a look and/or (preferably) check with the "log" module whether your UTF-8 is correctly passed into the target file.

AlexBGoode:
Sure. I did and found out that all my lovely UTF-8 messages are successfully appeared in mosquitto's log, in the log of mqttwarn in general (during pre-processing) and also in log-destination section:

*bash-3.2$ cat mqttwarn.log*
2015-11-13 16:35:11,525 DEBUG [mqttwarn] Message received on failover: **Привет 1**
2015-11-13 16:35:11,526 DEBUG [mqttwarn] Section [failover] matches message on failover. Processing...
2015-11-13 16:35:11,526 DEBUG [mqttwarn] Message on failover going to log:error
2015-11-13 16:35:11,526 DEBUG [mqttwarn] New `log:error' job: failover
2015-11-13 16:35:11,526 DEBUG [mqttwarn] Message on failover going to file:mqttwarn
2015-11-13 16:35:11,527 DEBUG [mqttwarn] Processor is handling: `log' for error
2015-11-13 16:35:11,527 DEBUG [mqttwarn] New `file:mqttwarn' job: failover
2015-11-13 16:35:11,528 DEBUG [log] * * * MODULE=services/log.pyc: service=log, target=error
[...]

PR #398, opened 2020-02-29
Encoding change needed to allow postgres inserts in python3

clarkspark:
I'm trying to use the postgres service on a fresh (pypi) install in a python3 virtual environment (after having used it for some time by cloning from github).  I could only get it to work without the forced encoding.  I also checked that the service still works in a python2 virtual environment as well.  I'm not sure what else this change might break.

I can provide more info detailing the specific messages that i was running into if it would help.


amotl:
Dear @clarkspark,

when working on the transition to Python3, we already fixed some charset encoding troubles (see #154, #360, #384, #395). So, I believe you did the right thing here and will just merge your updates right away.

Thanks a bunch for your contribution!

With kind regards,
Andreas.

PR #411, opened 2020-04-04
Fix #410: Load functions file at configuration load

### LINKED ISSUE #410: functions should be loaded once at startup time
Currently, the functions are loaded for every message. This is inefficient, and also prevents from being able to keep state. To keep state, I have to read and write a file every message. If the the functions were loaded once at startup, state could be kept in a module variable.

fhriley:
This has several advantages over loading the file at every message:

1) More efficient.
2) A bad functions file fails fast, i.e. mqttwarn will fail to start rather than waiting until a message arrives to show the failure.
3) When an error in the functions file exists, mqttwarn will fail with the actual exception string causing the failure, which makes it considerably easier to find and fix the problem.
4) State: since the functions module is only loaded once, the user can now keep state between function calls by using module level variables.

### DIFF
--- a/mqttwarn/configuration.py
+++ b/mqttwarn/configuration.py
@@ -13,6 +13,8 @@ try:
 except ImportError:
     HAVE_TLS = False
 
+from mqttwarn.util import load_functions
+
 logger = logging.getLogger(__name__)
 
 
@@ -80,6 +82,8 @@ class Config(RawConfigParser):
                 self.tls_version = ssl.PROTOCOL_SSLv3
 
         self.loglevelnumber = self.level2number(self.loglevel)
+        self.functions = load_functions(self.functions)
+
 
     def level2number(self, level):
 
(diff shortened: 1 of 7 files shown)


amotl:
Dear Frank,

thanks a bunch for your contribution. If @jpmens has no objections, I will be happy to merge right away and cut a new release.

With kind regards,
Andreas.

jpmens:
Merged, @fhriley, thank you very much!

fhriley:
I have the test fixed on python 2.7. How would you like to proceed? Submit a new pull request or revert this one?

jpmens:
I think a new PR would be best for us, thank you.

amotl:
> I have the test fixed on python 2.7.

Thanks!

> How would you like to proceed? Submit a new pull request or revert this one?

Just adding another PR will be perfectly fine!

PR #414, opened 2020-04-13
Fix for mqttwarn.service

trantor:
MQTTWARNINI is not present in the environment of the python process therefore mqttwarn doesn't find the configuration file.
`PassEnvironment` remedies that.
Also the variables are not expanded properly in the ExecStart.
If MQTTWARN_OPTIONS is an empty string it's passed as first argument to mqttwarn, resulting in the usage info being printed out.
I haven't checked but redirections shouldn't work in the ExecStart string.
Calling with an exec from sh -c remedies all that.
Tested on systemd 229 on Ubuntu 16.04.

PS I am not sure that the `--create-home` switch is needed in the `useradd` invocation of the commented guide.

### DIFF
--- a/etc/mqttwarn.service
+++ b/etc/mqttwarn.service
@@ -63,8 +63,9 @@ LimitNOFILE=65536
 Environment='STDOUT=/var/log/mqttwarn/mqttwarn.log'
 Environment='STDERR=/var/log/mqttwarn/mqttwarn.log'
 EnvironmentFile=/etc/default/mqttwarn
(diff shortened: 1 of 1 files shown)


amotl:
Dear Fulvio,

thanks for your contribution! As you have tested it successfully on an Ubuntu, I believe it will also still work on a vanilla Debian?

So, if @jpmens doesn't have any objections, we will probably merge your change right away for the better.

With kind regards,
Andreas.

jpmens:
Merged, thank you.

PR #418, opened 2020-04-30
fix indent

dlangille:
*** Sorry: IndentationError: unexpected indent (bad_funcs.py, line 6)

### DIFF
--- a/tests/bad_funcs.py
+++ b/tests/bad_funcs.py
@@ -3,4 +3,4 @@
 
 def foobar():
    foo # intentional indentation error
-    return True
+   return True


jpmens:
This is a test with an "intentional indentation error in it"

amotl:
Dear Dan,

thanks again for spotting this. In this case, it is really like @jpmens outlined above: I've put in this anomaly on purpose in order to test against a malformed custom functions file.

You can inspect the corresponding test case here:
https://github.com/jpmens/mqttwarn/blob/0.15.0/tests/test_util.py#L113-L115

Thanks already and with kind regards,
Andreas.

PR #420, opened 2020-05-03
Non-JSON payload should not generate warning

mortenf:
Plain text (and other formats) payloads should be OK and not generate a warning.


jpmens:
I agree, yes; and should have removed that years ago. Thank you. Merged.

PR #421, opened 2020-05-03
Fix missing namespace

mortenf:
A missing namespace on SerialException leads to an extra exception in case of errors.

(And extraneous white space a few lines above.)


jpmens:
Merged, thank you.

Issue #416, opened 2020-04-17
Update Dockerfile for Python 3


volschin:
Python 2.7 is retired. So Dockerfile should be updated to use Python 3 instead.

amotl:
That's true. Thanks for the heads up, @volschin. I will try to find some time to care about this detail. Otherwise, we are also happy to accept pull requests into that direction.

amotl:
Python 3 support within Docker will be arriving through #423, thanks to @koenvervloesem!

PR #422, opened 2020-05-04
Fix Dockerfile to use mqttwarn pip module

koenvervloesem:
The current Dockerfile doesn't work anymore.

This PR fixes the Dockerfile so it:

- installs the mqttwarn pip module (which pulls in the dependencies)
- runs the installed module instead of mqttwarn.py
- doesn't copy the whole repo directory into /opt/mqttwarn

I must admit that I'm not a poweruser of mqttwarn, so I haven't tested all functionality, but my tests of this Docker image with the smtp and log services on a Raspberry Pi 4 worked.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM python:2.7
 
 # based on https://github.com/pfichtner/docker-mqttwarn
 
-# install python libraries (TODO: any others?)
-RUN pip install paho-mqtt requests jinja2
+# install mqttwarn
+RUN pip install mqttwarn
 
 # build /opt/mqttwarn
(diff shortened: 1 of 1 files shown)


jpmens:
This looks reasonable to me and I've verified it installs correctly. Merging.
Thank you for the contribution (and good luck with the book!! :-)

amotl:
Dear Koen,

thanks a bunch for your contribution. It would also be cool to switch to Python 3 here, see #416. I would prefer `python:3.8.2-slim-buster` [1] if nobody has any objections.

With kind regards,
Andreas.

[1] https://hub.docker.com/_/python

PR #426, opened 2020-05-25
Swap use of encode in serial module

mortenf:
While I'm not quite certain of the reasoning behind the type detection code in pySerial, it seems that encoding the port name as bytes is not working, and keeping it unchanged works.

Likewise, but the other way around, writing a string to a serial port is rejected by pySerial, so here encoding is needed.

These observations may be specific to newer versions/combinations of mqttwarn, pySerial and/or Python 3.7, so it might be a good idea to have more eyes/experience on this.

Python: 3.7.3
pySerial: 3.4

### DIFF
--- a/mqttwarn/services/serial.py
+++ b/mqttwarn/services/serial.py
@@ -22,7 +22,7 @@ def plugin(srv, item):
     # While it may contain more than one item (e.g. pushover)
     # the `serial' service carries one two, i.e. a com name and baudrate
     try:
-        comName = item.addrs[0].format(**item.data).encode('utf-8')
+        comName = item.addrs[0].format(**item.data)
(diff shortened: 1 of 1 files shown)


amotl:
Dear Morten,

@zsquareplusc said within https://github.com/pyserial/pyserial/issues/7#issuecomment-142460713:

> The latest release should now raise a dedicated exception when Strings are written instead of bytes, which should make it easier to find such an error.

I will be happy to merge these updates right away and eventually welcome other people complaining about compatibility with Python 2. However, since Python 2 is officially deprecated, let's just move forward, right?

Any objections, @jpmens?

With kind regards,
Andreas.

jpmens:
@amotl no objections.

PR #427, opened 2020-06-05
Add missing text encode in pipe module

mortenf:
Another case of a missing text encode, that otherwise triggers an error:
`Cannot write to pipe: a bytes-like object is required, not 'str'`


amotl:
Thanks a bunch!

PR #428, opened 2020-06-05
Fix removal of as_user option within Slack plugin

mortenf:
When the optional parameter "as_user" is present, the Slack module attempts to remove it, in accordance with comments. It however fails to do so, as the shortened parameter list is assigned to a new/local variable, which is never used, instead of overwriting the original, as presumably intended.

This leads to failed logic later, when the number of parameters do not match expectations, which in turn leads to failed auth with Slack.


amotl:
Thanks for figuring this out! Merging.

PR #430, opened 2020-06-10
influxdb: make use of https

dkeipp:
Added config option or influxdb plugin "infxproto" to chose http or https.
default value is https to ensure backward compatibility.

### DIFF
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -1137,6 +1137,9 @@ Following is an ini example, showing the various connection properties for the I
 [config:influxdb]
 host      = 'influxdbhost'
 port      = 8086
+# Protocol for Influx connection [http|https], optional - default is set to http
+infxproto = 'https'
+
 username  = 'username'
 password  = 'password'
 database  = 'mqttwarn'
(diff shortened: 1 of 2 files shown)


jpmens:
Merged, thank you.

amotl:
Dear @dkeipp,

thanks for your contribution. I have a minor nit to address here: Could we rename `infxproto` to `scheme` here? I believe the former sounds a bit awkward.

With kind regards,
Andreas.

dkeipp:
> Dear @dkeipp,
> 
> thanks for your contribution. I have a minor nit to address here: Could we rename `infxproto` to `scheme` here? I believe the former sounds a bit awkward.
> 
> With kind regards,
> Andreas.

I have no problem to rename this parameter. I chose 'infxproto' because 'protocol' is used by the mqtt client section. This may cause misunderstandings among users.
Now the pull request is merged, and in my opinion we shouldnt change the parameter name.

Regards
Dennis

amotl:
Hi Dennis,

I took the chance that this has not settled within a release to apply the renaming as outlined above. After that, I've released `mqttwarn-0.16.2` so your improvements to the module should now be also available from PyPI.

Thanks again!

With kind regards,
Andreas.

Issue #391, opened 2019-12-17
Upgrade the service plugin for connecting to Azure IoT ("iothub")


amotl:
The `iothub` service plugin is currently defunct as `iothub-client==1.1.2.0` is deprecated. This plugin should probably be renamed to `azure-iot` and based on the _Azure IoT SDKs for Python_ available from https://pypi.org/project/azure/.

mortenf:
I will look into this over the coming weeks.

amotl:
Thanks to @mortenf, we can close this issue.

PR #433, opened 2020-07-04
Replaced iothub service with azure-iot, using MQTT, no dependencies

mortenf:
As per #391, the old "iothub" service is defunct. This change provides a new service, "azure-iot", using the MQTT protocol and thus without dependencies.


amotl:
Dear Morten,

thanks a bunch for your contribution!

I have just a minor remark: Can I humbly ask you to name the module `azure_iot.py`? While it currently works to use dashes in module names, I am aiming to simplify the plugin loading at some time in the future where this feature might be dropped.

With kind regards,
Andreas.

mqttwarn/services/azure-iot.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # iot hub name is stored in config
+    iothubname = item.config['iothubname']
+
+    # connect...
+    params = {
+        'hostname': iothubname + ".azure-devices.net",
+        'port': 8883,
+        'protocol': mqttclient.MQTTv311,
+        'qos': 0,
```
amotl, line 28: Is it possible to use an elevated QoS value here? If so, we might want to make configurable:
```
'qos': int(item.config.get('qos', 0))
```

mortenf: Sure. QoS 2 is not allowed, however, so I've added a check.

amotl:
This has been merged through #448. Thanks a bunch, @mortenf!

Issue #429, opened 2020-06-08
Populating InfluxDB with individual tags and fields from JSON message


psyciknz:
Been playing with fail2ban over the past couple of days....and from it managed to get a json formatted mqtt message out on each ban.
```
{ "ip": "37.49.226.155", "failures": "28", "jail": "ssh", "time": "1591645309.58". "country": "<country"> }
```
So now I want to get these into influx.

Here is my section.....I've tried all sorts of combinations.  I basically want a table, with tags for the jail (ssh) and ip, with values of the number of failures.

```
; -----------------------------------------
;             Fail2ban
; -----------------------------------------
[fail2ban/ban]
#topic = fail2ban/ban
targets = influxdb:fail2ban
format  = jail={jail},ip={ip} value={failures}
```

But nothing I do seems to affect the format.  The logs just try sending the entire json string as the value:
```
2020-06-09 07:41:50,985 DEBUG [connectionpool] http://influxdb:8096 "POST /write?db=network&rp=default&precision=ns HTTP/1.1" 400 219
2020-06-09 07:41:50,987 WARNING [influxdb] POST request failed: (400) {"error":"unable to parse 'fail2ban,topic=fail2ban_ban value={ \"ip\": \"37.49.226.155\", \"failures\": \"28\", \"jail\": \"ssh\", \"time\": \"1591645309.58\". \"country\": \"\u003ccountry\"\u003e }': invalid boolean"}
```

amotl:
Dear @psyciknz,

thanks for using mqttwarn and for writing in.

When looking at the configuration snippet [1] and the corresponding source code of the InfluxDB plugin [2], the `format` setting might suggest that it is possible what you are aiming at:
```
format = room=basement,entity=sensor2 temperature={payload}
```

However, it looks like the plugin currently is **not** capable of interpolating different fields from a JSON message received over MQTT into individual components of a signal to InfluxDB like you are intending here:
```
format  = jail={jail},ip={ip} value={failures}
```

With kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/blob/master/HANDBOOK.md#influxdb
[2] https://github.com/jpmens/mqttwarn/blob/0.16.1/mqttwarn/services/influxdb.py

psyciknz:
So I'd have to pull it apart the payload by a function - but even then I don't this the formatting would be as I wanted, or alter the influxdb service to allow any sort of tagging the user wants.

amotl:
[...]

PR #434, opened 2020-07-05
Load plugin service from any Python package

amotl:
Hi there,

service plugins can now be loaded from arbitrary Python packages with dotted string notation like `foo.bar.baz.qux`. This actually might implement the same thing originally intended by @psyciknz through #375.

Contrary to @psyciknz' implementation around the idea regarding
> The user can also specify a local directory and local services to launch.

this implementation doesn't mess with additional/optional directories with the need for specific code at all but instead builds upon native mechanisms from Python itself to let the user load service plugins from arbitrary Python packages. In this way, it is way more convenient for users to bundle their services within different/custom Python packages and deploy them alongside mqttwarn.

While I've retained the current code path for loading modules from within `mqttwarn.services` for now, this code can well be replaced by the new mechanics some time in the future, thus providing both kinds of module loading using the very same routines. The test `test_load_module_by_name_good` loading the built-in module `mqttwarn.services.file` using dotted string notation already reflects that.

With kind regards,
Andreas.

cc @rgitzel

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,7 +5,9 @@ mqttwarn changelog
 
 in progress
 ===========
+- Use ``allow_dirty = False`` within ``.bumpversion.cfg``
 - Use Python3 to create virtualenv
+- Bump version numbers for release tools packages
 
 
 2020-08-31 0.17.0
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,9 +5,10 @@ mqttwarn changelog
 
 in progress
 ===========
+- Use Python3 to create virtualenv
 
 
-2020-08-31 0.16.3
+2020-08-31 0.17.0
 =================
 - srv.mqttc is None when calling into a custom function. Thanks, Ben.
 - sundry changes for FreeBSD package. Thanks, Dan.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,12 +8,14 @@ in progress
 - Use ``allow_dirty = False`` within ``.bumpversion.cfg``
 - Use Python3 to create virtualenv
 - Bump version numbers for release tools packages
+- Add external plugin module loading. Thanks, @psyciknz!
 
 
 2020-08-31 0.17.0
 =================
 - srv.mqttc is None when calling into a custom function. Thanks, Ben.
 - sundry changes for FreeBSD package. Thanks, Dan.
+- Fix ``ZabbixSender.py``. Thanks, Ben!
 - service tweaks: nsca, zabbix
 
 2020-06-06 0.16.2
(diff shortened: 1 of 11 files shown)


amotl:
Dear @psyciknz,

I will be happy to see you building upon this implementation, maybe you will be able to add some creativity for getting this combined with Docker somehow?

> For docker, what I'll need to work out, is how any extra libraries might need to be loaded (probably a requirements file....maybe a dynamically loaded requirements/pip install when running).

While I admit there is some effort required, I also second that very much and recommend putting additional/custom modules into dedicated repositories on GitHub with appropriate names like `mqttwarn-plugin-foobar` which can be complemented by respective `setup.py`/`pyproject.toml` files and optionally even uploaded to PyPI. That way, getting everything installed (on Docker or elsewhere) is just determined by respective `pip install` commands without further ado.

With kind regards,
Andreas.

psyciknz:
> Dear @psyciknz,
> 
> I will be happy to see you building upon this implementation, maybe you will be able to add some creativity for getting this combined with Docker somehow?
> 
> > For docker, what I'll need to work out, is how any extra libraries might need to be loaded (probably a requirements file....maybe a dynamically loaded requirements/pip install when running).
> 
> While I admit there is some effort required, I also second that very much and recommend putting additional/custom service plugin modules into dedicated repositories on GitHub or GitLab with appropriate names like `mqttwarn-plugin-foobar`. They can be complemented by respective `setup.py`/`pyproject.toml` files and optionally even uploaded to PyPI.
> 
> That way, getting everything installed (on Docker or elsewhere) is just determined by respective `pip install` commands without further ado.
> 
> With kind regards,
> Andreas.

I think the easiest way to do it, it by changing the entry point of Docker from python mqttwarn to a run.sh in a mapped volume.

Then the user can change it to add requirements or pip command as required.

Actually, or just expose a requirements.txt that’s run always.

amotl:
Dear @psyciknz,

thanks for your quick answer.

Essentially, I just came across your comment from about a month ago
> I should merge you latest code into my branch for updates. Since my PR would be well out of date now. (#375)
>
> -- https://github.com/jpmens/mqttwarn/issues/429#issuecomment-640914683

and just thought "how hard could it be?".

So, may I assume you are satisfied with these updates which would give you the essentially the same feature to hook custom service plugins into mqttwarn?

With kind regards,
Andreas.

psyciknz:
@amotl  fine by me. I can always stick with my own version and we can work together later if there’s something about it I need adjusted.

I’ll look over the next couple of days at pulling your version and see if it’s works for me.

Do you have a dockerfile as well?  As I think I had my own image I’d built.

jpmens:
This looks very good to me, @amotl. Feel free to merge when you're ready.

PR #447, opened 2020-08-28
fix #446

### LINKED ISSUE #446: srv.mqttc is None when calling into a custom function
I have a custom format function which is designed to decompose an Owntracks location message and publish out some additional event messages (on different topics) for my home automation system to use.

This function uses `src.mqttc` to publish these event messages, but after migrating to docker based mqttwarn (latest) my function is failing with; 

```
WARNING [core] Cannot invoke owntracks_location(): 'NoneType' object has no attribute 'publish'
```

sumnerboy12:
I think this should resolve the issue I reported in #446. I have tested locally and my functions now have access to `mqttc` as expected.

### DIFF
--- a/mqttwarn/core.py
+++ b/mqttwarn/core.py
@@ -592,6 +592,9 @@ def connect():
         logger.error("Cannot connect to MQTT broker at %s:%d: %s" % (cf.hostname, int(cf.port), e))
         sys.exit(2)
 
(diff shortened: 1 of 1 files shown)


jpmens (approved):

jpmens:
Merged, thank you!

PR #450, opened 2020-08-31
[xbmc] fix encoding of data for python3

sumnerboy12:
I think all services that use `urllib` probably need reviewing also.


sumnerboy12:
Similar to #415

jpmens (approved):

jpmens:
Merged, thanks.

PR #451, opened 2020-10-03
Link to Supported Notification Services section

clach04:
Quicker access to list of supported notification mechanisms.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,7 +40,7 @@ sophisticated transformations.
 
 It comes with over **70 notification handler plugins** for a wide
 range of notification services and is very open to further contributions.
-You can enjoy the alphabetical list of plugins within the handbook_.
+You can enjoy the alphabetical list of plugins within the handbook_services_.
 
 A picture says a thousand words.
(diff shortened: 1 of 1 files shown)


jpmens:
I've labeled this as spam. Tell me it is not Hacktober spam.

clach04:
> I've labeled this as spam. Tell me it is not Hacktoberfest spam.

This is not Hacktoberfest spam. :-)

I'm really lazy and like having the clickable links closer so I don't have to scroll as much. You have a nice long list of backends so this make it easier to see them and jump directly to them.

I've opened some more serious bugs and PRs though, the broken/missing py3 support was unexpected in the two backends I tested. I would like to use upstream rather than my fork so if you get chance to review those that would be great. It would mean I could start to retire some of my hacky mqtt scripts and replace with mqttwarn :-)

jpmens:
@clach04 I saw you open serious PRs last night, but couldn't comment. Thank you for your contributions, and keep them coming!

Merged.

PR #458, opened 2020-10-06
:loud_sound: Initial Chromecast TTS support :loud_sound:

clach04:
Tested with Python 3.5.3.

Known to **not** work with Python 2.7 as current version of pychromecast no longer supports Python 2.7.


clach04:
See https://github.com/clach04/mqttwarn/blob/chromecast/HANDBOOK.md#chromecast for docs

jpmens:
I think until now we've been supporting Python 2.7, but as far as I'm concerned we can start dropping that, so I'm fine with this supporting Python 3 only.

jpmens:
Thank you very much; merged.

PR #460, opened 2020-10-07
:loud_sound: Amazon Alexa Say/Announce example :loud_sound:

clach04:
Sample for TTS to Alexa devices; Speaker, FireTV, etc.
Supports single devices and group announcements.


jpmens:
Merged. Thank you.

PR #467, opened 2020-11-20
Replace xmpppy by slixmpp

diorf:
Updated the xmpp plugin to use the slixmpp library instead of xmpppy.
xmpppy is not compatible with recent Python version and was preventing the plugin from working on my computer.

This will work as a drop-in replacement and will reuse the same configuration keys.

**Added dependency:** slixmpp https://lab.louiz.org/poezio/slixmpp
**Removed dependency:** xmpppy

### DIFF
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -2689,7 +2689,7 @@ recipients get the message.
 
 Requires:
 * XMPP (Jabber) accounts (at least one for the sender and one for the recipient)
-* [xmpppy](http://xmpppy.sourceforge.net)
+* [slixmpp](https://lab.louiz.org/poezio/slixmpp)
 
 ### `xively`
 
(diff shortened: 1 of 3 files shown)


amotl:
Dear @diorf,

thanks a bunch for your contribution. Using slixmpp will make mqttwarn require Python 3.7+. Currently, we still serve Python 2.7, 3.5 and 3.6. Maybe we should keep xmpppy for users who are eventually still on Python 2.7 and put another service side by side, which uses the slixmpp module?

With kind regards,
Andreas.

diorf:
Thanks @amotl 
I can send the request with an additional service to keep xmpppy & slixmpp in mqttwarn.

I'm pretty new to the world or PR: should I create a new one or update this one?
If the latter, I'll have to check how to do it.

Thanks,
Rémi

amotl:
Dear @diorf,

thanks for asking. Well, in this case I believe it will be better to just close this PR and open another one separately.

With kind regards,
Andreas.

diorf:
Closing this one as it was superseeded by #468

Issue #469, opened 2020-12-04
mysql error caused by encode('utf-8')


filippz:
Hello,

I've encountered an issue with mysql plugin when used with Python 3.8.5, where MySQL would throw an error:
` Cannot add mysql row: (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'b'mqttwarn'' at line 1")`

It seems to be caused by encode('utf-8') usage:
https://github.com/jpmens/mqttwarn/blob/37c7759bdbdc2de3916d978d73d430fc5765409d/mqttwarn/services/mysql.py#L48

As far as I can tell, those encode('utf-8') become byte strings in Python 3 and not unicode strings causing them to be prefixed with 'b' when used with %s formatting. For my use case it was enough to simply remove all encode('utf-8'), but I'm not sure what would be the proper way of fixing this?

Thanks,

amotl:
Dear Filip,

thanks a bunch for reporting this. Will you be able to submit a pull request in order to bring in this fix?

With kind regards,
Andreas.

PR #470, opened 2020-12-04
[mysql] Fix unicode vs byte issue for Python 3

filippz:
Fixes https://github.com/jpmens/mqttwarn/issues/469


amotl:
Thanks Filip!

PR #472, opened 2020-12-11
http service fix to send whole message - fixes #471

### LINKED ISSUE #471: Can http post send the whole message as is?
I tried it with;
```
[config:http]
timeout = 60
targets = {
          #method     #URL               # query params or None                              # list auth # Json
          'mymethod'    : [ "post", "https://example.com/post, None, None, True ]
          }

```
Based on the [code](
https://github.com/jpmens/mqttwarn/blob/1abee3d73ceabde04069eff5c6a385011e1833d1/mqttwarn/services/http.py#L101) I think this should be at least sth like;
```
            if params is not None:
                if tojson is not None:
                    encoded_params = json.dumps(params)
                    request.add_header('Content-Type', 'application/json')
                else:
                    encoded_params = urllib.parse.urlencode(params)
            else:
                if tojson is not None:
                    encoded_params = item.payload
                    request.add_header('Content-Type', 'application/json')
                else:
                    encoded_params = message
```

Also I have an error `'Request' object has no attribute 'add_data'`. Which is probably due to some 'request' lib incompatibility.

If I will have more time I will try to PR this, but bcs of other errors in the project and bcs I have close to zero python knowledge I'm stucked right now.

tg44:


jpmens:
Merged, thank you!

tg44:
Wow this was fast! Thanks!

PR #423, opened 2020-05-04
Add Docker Compose file

koenvervloesem:
This PR:
- adds a Docker Compose file that (together with the updated Dockerfile) uses `/etc/mqttwarn` instead of `/opt/mqttwarn`.
- adds comments to the Docker Compose file to document how to use it.
- updates the Dockerfile base image to `python:3.8.2-slim-buster`.

Tested on a Raspberry Pi 4 with smtp and log services and custom functions.

See https://github.com/jpmens/mqttwarn/issues/341#issuecomment-623661119

Issue #341, opened 2018-10-27
Publish the official Docker image


rgitzel:
@jpmens it would seem from https://github.com/jpmens/mqttwarn/pull/339 that another user has had to create their own image in Docker Hub, one that works slightly differently from the image defined at https://github.com/jpmens/mqttwarn/blob/master/Dockerfile

Could you publish the "official" image?  If nothing else, it would make it easier to support issues from Docker users

jpmens:
Look: all by myself! :-) 

![jmbp-4843](https://user-images.githubusercontent.com/60706/47616025-8466cc80-dab7-11e8-81bf-6b08c9b01e96.png)

jpmens:
Thanks to your work on the Dockerfile, etc. of course!

```
Removing intermediate container 5b8190284989
---> e57c8a34345f
Successfully built e57c8a34345f
Successfully tagged jpmens/mqttwarn:latest
Pushing index.docker.io/jpmens/mqttwarn:latest...
Done!
Build finished
```

Can you test the result for us, @rgitzel ?

rgitzel:
Congrats!!! :-)

I'll give it a try when I'm not tethered to my phone. 

The next trick will be adding a version tag.  Did you publish from your command line, or using the automated build?  Neat trick, I haven't tried that one before.

rgitzel:
Unfortunately I didn't link to the manual instructions, here, that might have saved you some time:  https://github.com/jpmens/mqttwarn#build-the-image

I'm a bit leery of automatically publishing the image on a commit to the master branch, as it would update the `latest` image even for a simple typo fix in the README. 

The typical thing to do for Docker images is to explicitly give them a version tag (e.g. "1.7") as the associated applications are released.  Then the tag `latest` is applied to the most recent image (essentially acting as an alias to the version).   

Each version tag is considered immutable, but only by convention, as Dockerhub will let you move any tag to any image.  As code changes are made and the application version changes, new images are published with _new_ tags (e.g. "1.7.1", "1.8").  The existing tags all remain in place, except for `latest` which is moved to the image that should be considered the current release.

So... in this case, rather than republishing the image automatically on a Git commit, it is probably clearer to do the build/tag/push steps manually when a new version of the app is released.   Probably the Dockerhub automation could do that, but that might be more complicated to set up than it's worth?  When the `dev` version is ready, it could become part of the process that publishes the Pip artifact?  (i.e. push the image to Docker Hub, instead of Docker Hub pulling it from Git).

[...]

PR #473, opened 2020-12-11
fix the docker file

tg44:
As I tried to resolve a http plugin issue, I tried to modify/debug my container, and I found out some really strange things.

#### Documentation
The documentation refers the /opt dir while in #423 it modified to /etc (I reverted back to /opt)

#### Dockerhub vs versions
If you check the digests on [dockerhub](https://hub.docker.com/r/jpmens/mqttwarn/tags?page=1&ordering=last_updated) it is the same in the latest, 0.20.0, 0.19.0, ... 0.16.0 versions. Also if you `docker run -it --rm jpmens/mqttwarn:0.20.0 mqttwarn --version` you will get back 0.16.0

#### Dockerfile
In general I think it is a really bad idea to just pip install the tool... This should be something like in this PR.
The contents of the dockerfile should build from the source. Not from a pip install...

I think the 0.16.0 bug comes from the fact that docker caches image layers, so when sb pushed 0.16.0 he/she did run the docker build, got the layers cached, and bcs there was no copy command, next time the docker build found no changes, and happily published 0.20.0 as the same image as the 0.16.0 was.

#### Concerns about this PR
I have close to zero idea about the python world, so this setup.py call could be as bad as the pip install was before (but at least for different reasons). BUT this PR unlocks the option in dockerhub to maintain a develop/master tag with the dockerhub autobuild from master trigger thingy.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,21 @@
 FROM python:3.8.2-slim-buster
 
-# based on https://github.com/pfichtner/docker-mqttwarn
+#install dir
+RUN mkdir -p /app
 
-# install mqttwarn
-RUN pip install mqttwarn
-
-# create /etc/mqttwarn
-RUN mkdir -p /etc/mqttwarn
+# run dir
+RUN mkdir -p /etc/mqttwarn/
 WORKDIR /etc/mqttwarn
-
-# add user mqttwarn to image
-RUN groupadd -r mqttwarn && useradd -r -g mqttwarn mqttwarn
-RUN chown -R mqttwarn:mqttwarn /etc/mqttwarn
-
-# process run as mqttwarn user
-USER mqttwarn
 
 # conf file from host
 VOLUME ["/etc/mqttwarn"]
 
 # set conf path
 ENV MQTTWARNINI="/etc/mqttwarn/mqttwarn.ini"
+
+# copy and install requirements
+COPY . /app
+RUN cd /app && python setup.py develop
 
 # run process
 CMD mqttwarn
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -3357,14 +3357,14 @@ This is the typical way of running `mqttwarn`.
 
 From the folder containing your `mqttwarn.ini` file:
 
-```
+```console
 $ docker run -d --rm --name mqttwarn \
-    -v $PWD:/opt/mqttwarn/conf \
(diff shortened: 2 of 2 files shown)


koenvervloesem:
One question: why the move back to `/opt`? I moved the path to `/etc` because of @amotl's suggestion in https://github.com/jpmens/mqttwarn/issues/341#issuecomment-623606799.

tg44:
It was easier than finding the misusages in the documentation. Pure laziness :D We have no real reason for using either. I moved the whole app to the `/app` dir, we could use the `/conf` `/config` `/app/config` etc. dirs too... So no real reason.

jpmens:
one request please: if there’s “no real reason” then let’s not do it.

tg44:
Lets not do it what? I mean its now consecvent with the current state of the documentation. We either need to fix the docs or the directories in the docker image. I went with the second one.

When I said theres no real reason I ment that there is no real reason to use any of that. There was no real reason to modify it in the first case either.

I can change it back, but then we need to review the documentation more carefully.

tg44:
I force pushed it, also fixed the docs.

amotl:
Dear @tg44, @koenvervloesem and @jpmens,

instead of fixing the Docker file, I tried to give the Docker documentation some love with #493, also reusing some content from this patch (thank you very much, Gergő). Let me know what you think about it.

With kind regards,
Andreas.

amotl:
Dear @tg44,

when looking through your patch, I found you want to load service plugins from custom locations, right? #496 now improves the situation, you can find the corresponding documentation for loading external services at [1].

Please let us know what you think about it and if that satisfies your use case. If you see any further things to be improved, please open a separate issue.

With kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/blob/master/HANDBOOK.md#loading-external-services

tg44:
I don't think you can fix this with only the documentation... Please give time to read through both this thread, check my file modifications, and my response to #495. At least some part of this PR should be merged in, if you want working and up to date docker versions.

The main problem is not adding external services, that was just the tip of the iceberg. I saw that you pushed a 0.21 version which get a new digest, but you need to understand, that this happened only bcs you either changed computer, or removed the previously builder image from your machine. Also, docker hub has a pretty good github integration, which is only a few clicks to make the image automatically, I think you should set that up too.

amotl:
Thank you Gergő, you are right! We will address the issues and will be happy to continue the discussion at #495.

PR #483, opened 2021-02-14
New service launch

joerggollnick:
This adds `launch` functionality to mqttwarn.
It combines execute with flexible parameters and mqttpub afterwards.
Configuration is outgoing_topic, qos, retain, cmd, ...
It publishes stdout to outgoing_topic


joerggollnick:
to many commits / HANDBOOK.md missing

PR #496, opened 2021-06-02
Load service plugins from both modules and files

amotl:
Hi again,

this patch improves the situation to load service plugins from custom locations.

With kind regards,
Andreas.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ in progress
 ===========
 
 
-2021-06-02 0.21.0
+2021-06-03 0.21.0
 =================
 - [xmpp] Add slixmpp plugin and documentation. Thanks, Remi!
 - [mysql] Fix unicode vs byte issue for Python 3. Thanks, Filip!
@@ -18,6 +18,7 @@ in progress
   responds with MQTT publish. Thanks, Jörg!
 - [core] Remove "os.chdir" as it is apparently not needed anymore. Thanks, Dan!
 - [ci] Run tests on Python 3.9, remove testing on Python 3.5
+- [core] Load service plugins from both modules and files
 
 
 2020-10-20 0.20.0
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -23,9 +23,8 @@ I've written an introductory post, explaining [what mqttwarn can be used for](ht
     + [Custom functions](#custom-functions)
     + [Templates](#templates)
   * [Periodic tasks](#periodic-tasks)
-  * [Running with Docker](#running-with-docker)
-    + [Run the Image](#run-the-image)
-    + [Build the image](#build-the-image)
+  * [Running with Docker](#docker)
+  * [Loading external services](#loading-external-services)
   * [Examples](#examples)
     + [Low battery notifications](#low-battery-notifications)
     + [Producing JSON](#producing-json)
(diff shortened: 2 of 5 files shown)


psyciknz:
Reply here with some workings?
So in my mqttwarn folder that I use for my docker container I have:
```
mqttwarn.ini
funcs.py
local-services/smsgateway.py
```
When I run my container, I map this local folder to /etc/mqttwarn in the container.  That means that the config is at /etc/mqttwarn/mqttwarn.ini

Question, possibly for the container maintainer, and for the documentation, when you provide as example of a local service to : 
```
[config:tests.acme.foobar]
```
or
```
[config:tests/acme/foobar.py]
```
Where is that referenced from?  The /etc/mqttwarn folder, or is it based off where the mqttwarn.py program lives (I've not found that yet).

So any I've managed to get my "local service" being found with the notation:
```
[config:/etc/mqttwarn/local-services/smsgateway.py]
```
But that brings up another issue that I think I found in my original pull request - how to get packages used by the local service, into the container or python.
```
2021-06-09 19:40:52,866 ERROR [core] Unable to load service "/etc/mqttwarn/local-services/smsgateway.py" from file "/etc/mqttwarn/local-services/smsgateway.py": No module named 'httplib'
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/mqttwarn/core.py", line 567, in load_services
    service_plugins[service]['module'] = load_module_from_file(modulefile)
  File "/usr/local/lib/python3.9/site-packages/mqttwarn/util.py", line 164, in load_module_from_file
    return imp.load_source(digest, path, fp)
  File "/usr/local/lib/python3.9/imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 711, in _load
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 855, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/etc/mqttwarn/local-services/smsgateway.py", line 8, in <module>
    from httplib import HTTPConnection, HTTPException
ModuleNotFoundError: No module named 'httplib'
2021-06-09 19:40:52,868 CRITICAL [core] Unable to load service "/etc/mqttwarn/local-services/smsgateway.py"
```

I think I can docker run jpmens/mqttwarn pip install httplib and get over it...But I'm just trying to think of a way of doing that with the container being constructed rather than a manual operation you have to run on a new container instance.

With the local services pull request I had, I think I switched the run command one in the local host folder, or externalised a requirements.txt file that the user could then modify.

psyciknz:
Oh...acutally the first bit about the paths....I think it is the local folder.
I switch the mqttwarn config from /etc/mqttwarn/local-services/smsgateway.py to local-services/smsgateway.py and it found it....and then I'm back to the urllib problem.

psyciknz:
And the 2nd part while we might need something for installing packages, most of my issues relate to a change in python from 2.7 to 3.x so the libraries it could not find had been renamed.....

psyciknz:
Ok, and a test (invalid payload) but I saw my local service called smsgateway kick into gear and try and run:
```
2021-06-09 20:06:59,172 DEBUG [core] Message received on smsgateway/inbox: test
2021-06-09 20:06:59,172 DEBUG [core] Section [smsgateway/inbox] matches message on smsgateway/inbox. Processing...
2021-06-09 20:06:59,173 DEBUG [core] Cannot decode JSON object, payload=test: Expecting value: line 1 column 1 (char 0)
```

PR #498, opened 2021-06-03
Fix Docker setup

amotl:
Hi there,

@tg44 made us aware that we have been tripped by unwanted cache hits when building Docker images, thanks again. This patch hopefully mitigates the problem. What do you think, Gergő?

With kind regards,
Andreas.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,9 @@
+*
+!etc
+!mqttwarn
+!tests
+!setup.py
+!MANIFEST.in
+!*.ini
+!*.md
+!*.rst
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,14 +6,14 @@ mqttwarn changelog
 in progress
 ===========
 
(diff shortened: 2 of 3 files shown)


tg44:
This seems like a good docker file, the only thing that can be better to move down the copy+install right before the cmd so most of the layers can be reused, but it is really just nitpicking. (moving 10-12 to 30, so from 11 layers we could reuse 8 or so)

jpmens (approved):
If @tg44 is happy, I am happy. :-)

amotl:
Great. Can you check 0ea8f76d55b5b, @tg44?

tg44:
Seems ok to me!

jpmens:
rabbit-7989

```
Could not install packages due to an OSError: [Errno 13] Permission denied: '/home/mqttwarn'
Check the permissions.
Removing intermediate container e44affdb045a
The command '/bin/sh -c pip install /src' returned a non-zero code: 1
```

PR #490, opened 2021-06-02
Remove "os.chdir" as it is apparently not needed anymore

amotl:
Hi there,

will this fix #489 without causing any other trouble?

With kind regards,
Andreas.

/cc @dlangille

PR #499, opened 2021-06-03
Respect relative path names within configuration file

amotl:
Hi there,

after removing a call to `os.chdir()` with #490, I tried to dive in a bit and came up with an idea to _support loading auxiliary files with **relative** paths_. When doing so, they should be treated to use the directory of the `.ini` configuration file as the base directory. This applies to both the function file as well as external service module files.

The feature was implemented in a backward-compatible manner in order not to break existing configurations. Both scenarios are supported by respective tests multiplexing the current ones.

With kind regards,
Andreas.

/cc @psyciknz, @tg44

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,13 @@ mqttwarn changelog
 in progress
 ===========
 
+
+2021-06-03 0.22.0
+=================
+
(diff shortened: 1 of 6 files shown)


jpmens (approved):
good show!

PR #500, opened 2021-06-03
Rename "http.py" module to reduce conflicts with stdlib

### LINKED ISSUE #436: Conflict between Python 3 http module and http service
When trying to execute a service plugin file directly, as e.g. [suggested by the tootpaste service](https://github.com/jpmens/mqttwarn/blob/master/HANDBOOK.md#tootpaste), Python gets confused by the http service plugin file:

`ModuleNotFoundError: No module named 'http.client'; 'http' is not a package`

Apparently, the Python 3 package http is "overridden" by the http service plugin.

I'm not sure how this should be fixed - only real solution seems to be renaming the http service, which will surely break things out there.

Executing a service plugin file directly is useful for initialisation purposes, but also for development/testing.

amotl:
Hi there,

this patch renames the `http.py` module to `http_urllib.py` in order to reduce conflicts with stdlib. It will resolve #436.
For backward compatibility reasons, the module is still known by the same name `"http"`, so no configurations will break.

With kind regards,
Andreas.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,14 @@ in progress
 ===========
 
 
+2021-06-03 0.23.0
+=================
+
+- [http] Rename ``http.py`` module to ``http_urllib.py`` to reduce conflicts with stdlib.
+  For backward compatibility reasons, it is still available by the same name, so no
+  configurations will break.
+
+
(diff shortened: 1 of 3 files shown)


jpmens (approved):
LGTM, thanks

Issue #503, opened 2021-06-07
Docker and modules (eg slack_sdk)


psyciknz:
Trying to look at the official image.
As I'm using slack, I need to pip install slack_sdk.
At this stage, I'm just planning on running 
```docker run --it <volume> jpmens/mqttwarn pip install slack_sdk.```

But i get a permissions error:
```
WARNING: The directory '/home/mqttwarn/.cache/pip' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
Defaulting to user installation because normal site-packages is not writeable
Collecting slack_sdk
  Downloading slack_sdk-3.6.0-py2.py3-none-any.whl (225 kB)
     |████████████████████████████████| 225 kB 566 kB/s
Installing collected packages: slack-sdk
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/home/mqttwarn'
Check the permissions.
```

How have others done this?

**Comment by psyciknz** [comment]
I did try this as well, but also seems to get permissions error:
```
pip cache dir --cache-dir=/usr/local/lib/python3.9/site-packages/
WARNING: The directory '/usr/local/lib/python3.9/site-packages' or its parent directory is not owned or is not writable by the current user. The cache has been disabled. Check the permissions and owner of that directory. If executing pip with sudo, you should use sudo's -H flag.
ERROR: pip cache commands can not function since cache is disabled.
```

**Comment by amotl** [comment]
Dear @psyciknz,

thank you for describing your usage scenario and apologies it didn't work out of the box. As far as I know, a core concept of Docker images is that they are immutable, so in order to customize their contents, one would always need to build a custom image derived from another base image.

In order to outline a solution to your problem based on this, I just submitted #504.

With kind regards,
Andreas.

**Comment by amotl** [comment]
I am only now seeing that @sumnerboy12 apparently also added 7f287fc0, adjusting the main Dockerfile to also install the `slack-sdk` and `gpxpy` Python modules ;].

**Comment by sumnerboy12** [comment]
Whoops, did that go into master? Apologies, I was only meant to push that to my fork!

**Comment by psyciknz** [comment]
This might be a possible docker-compose usage....

```
services:
  mqttwarn:
    build: .
    container_name: mqttwarn-test
    hostname: mqttwarn
    environment:
      - ARCH=amd64
      - BASEPATH=/hdd/docker-data
    restart: always
[...]

PR #504, opened 2021-06-07
Add example Dockerfile outlining how to add additional Python modules

amotl:
Hi there,

this patch adds a `Dockerfile.mqttwarn-slack` file, based on the discussion started by @psyciknz at #503, so it will
install the Python Slack SDK into a custom Docker image called `mqttwarn-slack`.

With kind regards,
Andreas.

### DIFF
--- a/Dockerfile.mqttwarn-slack
+++ b/Dockerfile.mqttwarn-slack
@@ -1,0 +1,19 @@
+# Docker build file for mqttwarn, with Slack SDK.
+# Based on https://hub.docker.com/r/jpmens/mqttwarn.
+#
+# Invoke like:
+#
+#   docker build --tag=mqttwarn-slack --file=Dockerfile.mqttwarn-slack .
+#
+
+# Derive from upstream image.
+FROM jpmens/mqttwarn:0.23.0
+
+# Make package installation run as "root" user
+USER root
+
+# Install Slack SDK.
+RUN pip install mqttwarn[slack]
+
+# Make process run as "mqttwarn" user
+USER mqttwarn
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -131,3 +131,18 @@ The name `mqttwarn-local` is not meaningful, other than making it obvious when
 you run it that you are using your own personal image. You can use any name you
 like, but avoid `mqttwarn` otherwise it's easily confused with the official
 images.
+
+
+## Installing additional Python packages into Docker image
+
+In order to use `mqttwarn` with additional Python modules not included in the
+baseline image, you will need to build custom Docker images based on the
+canonical `jpmens/mqttwarn`.
+
+We prepared an example which outlines how this would work with the Slack SDK.
+By using the `Dockerfile.mqttwarn-slack`, this command will build a Docker
+image called `mqttwarn-slack`, which includes the Slack SDK:
+
+```shell
+docker build --tag=mqttwarn-slack --file=Dockerfile.mqttwarn-slack .
+```


Dockerfile.mqttwarn-slack
```diff
@@ -0,0 +1,19 @@
...
+#
+
+# Derive from upstream image.
+FROM jpmens/mqttwarn:0.23.0
+
+# Make package installation run as "root" user
+USER root
+
+# Install Slack SDK.
+RUN pip install mqttwarn[slack]
```
sumnerboy12, line 16: I am not aware of this syntax - can you explain how this works, and how you would write this to also install a second module - say `gpxpy`?

psyciknz: I thnk it's a custom dockerfile, I think I saw somewhere that is gpxpy is another mqttwarn library it'd be: RUN pip install mqttwarn[slack,gpxpy]

So would need to be my own build....I might see if I can docker run jpmens/mqttwarn pip install mqttwarn[slack] instead as a one off, so I don't need my own docker image, as to be that seems to default the purpose of using the generic.

sumnerboy12: yeah I meant the `pip install mqttwarn[<module>]` syntax - never seen that before.

sumnerboy12: https://stackoverflow.com/questions/46775346/what-do-square-brackets-mean-in-pip-install

sumnerboy12: Right - so the extra tag is referring to the modules defined in `setup.py` - so it wouldn't work by just adding `gpxpy` since that is included in `setup.py`.

amotl: Hi Ben and @psyciknz,

thanks for signaling interest on this patch.

> I am not aware of this syntax.

Some Python packages have optional [extras](https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/#installing-extras), and so does `mqttwarn` in its [setup.py at the "extras" section](https://github.com/jpmens/mqttwarn/blob/0.23.0/setup.py#L20-L117) for all of the dependencies of the built-in service plugins.

> How you would write this to also install a second module?

You would invoke it like
```shell
pip install mqttwarn[slack,twitter]
```

With kind regards,
Andreas.

sumnerboy12: Ok thanks @amotl - very nice.

sumnerboy12:
@amotl is it worth adding something to `.gitignore` so that users can safely add their custom `Dockerfile.xxx` and not be concerned with it being pushed to git?

psyciknz:
> @amotl is it worth adding something to `.gitignore` so that users can safely add their custom `Dockerfile.xxx` and not be concerned with it being pushed to git?

Surely that's only an issue if you're forking the main mqttwarn repo.

sumnerboy12:
> Surely that's only an issue if you're forking the main mqttwarn repo.

Well I can imagine a lot of people will just pull the main repo and work locally, to build custom docker images, this would just make it cleaner, without causing git to flag any changes.

But yes, it may well be unnecessary.

Dockerfile.mqttwarn-slack
sumnerboy12: BTW - just tested this for my use-case and it works very nicely - very handy!

psyciknz:
> > Surely that's only an issue if you're forking the main mqttwarn repo.
> 
> Well I can imagine a lot of people will just pull the main repo and work locally, to build custom docker images, this would just make it cleaner, without causing git to flag any changes.
> 
> But yes, it may well be unnecessary.

Yeah that's why I did it in the docker-compose file, the only thing I couldn't work out was how to change the name of the dockerfile as part of a compose build command....

But if you're writing your own dockerfile and using docker, there should be no need to fork the mqttwarn repo, as your docker file pulls from dockerhub.

amotl:
Thank you for your feedback. Let's improve this situation further and add additional convenience by implementing the features outlined at #506.

PR #511, opened 2021-06-09
[prowl] Update service plugin to use "pyprowl" instead of "prowlpy"

### LINKED ISSUE #507: Improve Prowl notification plugin
Hi there,

according to [1], the Prowl service plugin is based on @jacobb's `prowlpy` [2]. As the most recent release happened on 14 Mar 2011, we should probably start looking for a more recent wrapper. @toddrob99's `pyprowl` might come to the rescue.

With kind regards,
Andreas.

P.S.: This would currently break installation with `pip install mqttwarn[all]` for building the Docker image `mqttwarn-full`, added with #506, so we had to exclude this package there.

[1] https://github.com/jpmens/mqttwarn/blob/0.23.0/setup.py#L181
[2] https://github.com/jacobb/prowlpy
[3] https://github.com/toddrob99/pyprowl

amotl:
Hi there,

this patch might resolve #507 by updating the Prowl service plugin to use the `pyprowl` package instead of `prowlpy`. This code is currently completely untested. Is anyone in the community able to verify this before or after the merge?

With kind regards,
Andreas.

/cc @jacobb, @martinjo, @MansM

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ mqttwarn changelog
 in progress
 ===========
 
+- [prowl] Update service plugin to use "pyprowl" instead of "prowlpy"
 
 2021-06-08 0.23.1
 =================
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -2089,10 +2089,10 @@ targets = {
 ![Prowl](assets/prowl.jpg)
 
 Requires:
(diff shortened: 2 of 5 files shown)


jpmens (approved):
Let's merge this and get it in. If somebody tests later and finds a flaw we'll address it then.

PR #515, opened 2021-06-13
[ci] Build Docker images for riscv64 and mips64

amotl:
What the title says.


amotl:
This apparently does not work out of the box.

PR #520, opened 2021-06-15
Tests: Refactor configuration files, improve naming things

amotl:
Hi again,

this is just a cosmetic update in order keep us sane when maintaining and expanding the test suite. The patch will, effectively:

- Move all configuration files (.ini and .py) relevant for testing into the folder `tests/etc`.
- Adjust some bits on the "naming things" side in order to improve telling things apart.

With kind regards,
Andreas.


jpmens (approved):
LGTM

PR #522, opened 2021-06-17
Begin adding software tests for service plugins

amotl:
Hi there,

this patch starts adding software tests for all service plugins.

The tests will not be integration tests, all domain-specific outbound calls or libraries will be mocked instead. Modulo that detail, we aim for 100% test coverage. More work will have to go into this endeavor, currently only 11% of all plugins are being tested.

While adding some tests, minor issues with individual service plugins have been fixed. At the same time, this patch drops official support for Python 2.

With kind regards,
Andreas.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ["2.7", "3.6", "3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
         include:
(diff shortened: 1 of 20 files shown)


jpmens:
Great we're dropping Python 2; I welcome that. :-)

amotl:
Let's integrate it to make it easier for @psyciknz to work with this improvements, see https://github.com/jpmens/mqttwarn/issues/526#issuecomment-863649072 ff.

amotl:
> I will try to continue working on this topic after vacation.

Let me know if you want to bring this in beforehand. I would also cut a new release immediately after.

jpmens (approved):
LGTM

Issue #355, opened 2019-01-13
Docker Warning: Control your logging


psyciknz:
Not an issue, but a note worth looking at for those with Docker.

By default, docker does not manage it's own logs for each container.  And defaults to json-file.  When my system ground to a halt, I found the mqttwarn docker log (since I send to stream), sitting at 7.2gb.

The answer (for me), was to configure the container, or docker service to set a file size limit on the log file:
https://docs.docker.com/config/containers/logging/json-file/

```
{
  "log-driver": "json-file",
  "log-opts": {
    "max-size": "10m",
    "max-file": "3"
  }
}
```

jwyse:
Thanks for sharing this.  I just freed about 9 GB of logs from other (not mqttwarn) containers.

I initially overlooked this part, which led to some head-scratching: (emphasis mine)

> Restart Docker for the changes to take effect for newly created containers. **Existing containers do not use the new logging configuration.**

So you have to restart the docker service AND recreate the container, for this to go into effect.

amotl:
Dear @psyciknz,

also thanks for sharing this. Do you see any chance to submit a patch to add this to the documentation at [1] using appropriate command line options like `--log-opt max-size=10m`?

With kind regards,
Andreas.

[1] https://github.com/jpmens/mqttwarn/blob/main/DOCKER.md

PR #529, opened 2021-06-19
Support for VSCode, Improve Docker documentation, Add "publishversion" feature

psyciknz:
- Updated the Docker.md with information for docker logging, see #355.
- Added a new configuration option in for publishversion and a topic, upon connecting to mosquitto, it will publish the current mqttwarn version.  Perhaps could have been more elegant ways of doing it via a function, happy to have that removed from the merge.
- Also added some specific files for running with VSCode, since it was difficult to determine how to run as a module.  These exist in the .vscode directory with a launch.json file.  They do require a run.ini file.
- I do note some old commits  listed here for a cloudflare service I had (services/cloudflare.py), but I re-forked this repository so do not expect them in there....please verify I'm not adding them by accident.

EDIT: I note the original reason I started the branch to fix pushover, has already been fixed in the main branch,


DOCKER.md
```diff
@@ -107,6 +107,31 @@ Add this argument to `docker run`:
...
+```yml
+logging:
+  driver: "json-file"
+  options:
+    max-size: 10m
+    max-file: "3"
+    env: "os"
+```
+
+
```
amotl, line 134: Can you also add some information about how to adjust the logging options _on the command line_ as outlined within https://github.com/jpmens/mqttwarn/issues/355#issuecomment-863470905?

tests/etc/full.ini
```diff
@@ -57,6 +57,9 @@ functions = 'tests/etc/functions_good.py'
 ; name the service providers you will be using.
 launch    = log, file, tests.acme.foobar, tests/acme/foobar.py
 
+; Publish the current mqttwarn version to the topic (retained)
+publishversion = mqttwarn/version
+
```
amotl, line 62: Can you add this configuration option to `mqttwarn/examples/basic/mqttwarn.ini` instead?

mqttwarn/core.py
```diff
@@ -627,6 +628,10 @@ def connect():
     # Update our runtime context (used by functions etc) now we have a connected MQTT client
     context.invoker.srv.mqttc = mqttc
 
+    # if publishversion configuration entry exists, use it to publish the current mqttwarn version
+    if cf.publishversion:
```
amotl, line 632: Did you make sure mqttwarn does not croak when accessing `cf.publishversion` without that configuration item being defined in `mqttwarn.ini`?

tests/etc/full.ini
amotl: Can you add a corresponding section about that option to the documentation in `HANDBOOK.md`?

amotl (changes requested):
Dear @psyciknz,

thanks a stack for this contribution, I like its ingredients very much. However, can I humbly ask you to submit separate patches for individual features or improvements and also try to keep a reasonable commit history?

As you might have noticed, I tried to support you on this matter by addressing some of the improvements using separate pull requests, also addressing other things in your area. Specifically, #530 adds an appropriate launch configuration for VSCode.

Also, I added some inline comments to give you early feedback on different details. Thank you again for working on this!

With kind regards,
Andreas.

tests/etc/full.ini
psyciknz: Thanks, wasn't sure where I was to add it, and the only config I saw that was "full" was in this example....

mqttwarn/core.py
psyciknz: actually no, now that I think about it, it will cause an exception.  As publishversion will not exist and not return a none just because it's not there.

amotl:
Dear @psyciknz,

thanks again for your contributions. After addressing all your improvements with those separate patches listed below, I believe it is safe to close this PR.

- #539
- #552
- #577
 
If you find that anything is still not right yet, we will be happy to receive further patches.

With kind regards,
Andreas.

Issue #516, opened 2021-06-13
Running mqttwarn in vscode.


psyciknz:
moved to official image as you know, and have found one of my items not working (in the pushover service I think there's a 2.7 > 3.9 incompatibility).

I think i have a fix for it, but just wanting to run it in debug mode in visual studio code.  The structure has changed significantly since I last ran it.

What python file do I execute?
And where should the mqttwarn.ini file reside to run?

amotl:
Dear @psyciknz,

thank you for your report.

> I found one of my items not working: In the pushover service I think there's a 2.7 > 3.9 incompatibility.

That could well have happened, apologies. We made the transition to Python 3 the other day but haven't been able to keep up with migrating and verifying all service plugins and just take care about them as we go. We would appreciate any patches for the pushover service plugin and beyond.

> What python file do I execute?

The main program file entrypoint is just called `mqttwarn.py`, it will invoke `mqttwarn.commands:run`.

> Where should the `mqttwarn.ini` file reside to run?

By default, it will use a file `mqttwarn.ini` from the current working directory. Otherwise, you can adjust the `MQTTWARNINI` environment variable to point to a different configuration file.

With kind regards,
Andreas.

psyciknz:
> > What python file do I execute?
> 
> The main program file entrypoint is just called `mqttwarn`, it will invoke `mqttwarn.commands:run`.
> 
If you were running it on linux, what python command (to what file) would you execute:
python mqttwarn/core.py?    AS I don't see a mqttwarn.py file in the file system any more.


I think I've manage to get around it by creating a run.py module that calls mqttwarn.commands.run().

amotl:
Hi @psyciknz,

> If you were running it on linux, what python command (to what file) would you execute?

The command entrypoint `mqttwarn` is actually a Python file, it just does not have a `.py` extension. When installing the package in development mode into a virtualenv, the `mqttwarn` program will be located within its `bin/` directory (see below).

I hope VSC will be able to invoke that program somehow?

With kind regards,
Andreas.

---

```shell
$ cat .venv/bin/mqttwarn
#!/Users/amo/dev/daq-tools/sources/mqttwarn/.venv/bin/python3.9
# EASY-INSTALL-ENTRY-SCRIPT: 'mqttwarn','console_scripts','mqttwarn'
import re
import sys

# for compatibility with easy_install; see #2198
__requires__ = 'mqttwarn'

try:
    from importlib.metadata import distribution
except ImportError:
[...]

PR #530, opened 2021-06-19
Add launch configuration for VSCode

amotl:
Hi there,

we are coming from #516 and #529 here. This patch will make development on the codebase of mqttwarn using VSCode work without further ado, probably in a platform-agnostic manner.

With kind regards,
Andreas.

/cc @psyciknz


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
+    // VSCode launch configuration settings file for mqttwarn.
```
jpmens, line 2: I shall assume comments are permitted in this file specifically for VScode.

jpmens (approved):

amotl:
Dear @psyciknz,

please let me know if you can satisfy VSCode with this patch.

May I also humbly ask which operating system you are using? On the detail of how to setup the virtualenv stuff, we might have to adjust the documentation when it comes to, for example, running that procedure on Windows. In that case, I am kindly asking for corresponding contributions to the guidelines.

With kind regards,
Andreas.

.vscode/launch.json
amotl: That was new to me as well, but right - nothing croaks on this end. It is probably already using JSON5?

PR #535, opened 2021-06-19
Make parts of the test infrastructure reusable

amotl:
Hi again,

this refactors parts of the testing code to make it more reusable. In that manner, the relevant code has not to be duplicated into "mqttwarn-contrib".

With kind regards,
Andreas.

### DIFF
--- a/mqttwarn/model.py
+++ b/mqttwarn/model.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# (c) 2021 The mqttwarn developers
+from dataclasses import dataclass
+from typing import Dict, List, Union
+
+
+@dataclass
+class ProcessorItem:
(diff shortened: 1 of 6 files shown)


mqttwarn/model.py
```diff
@@ -0,0 +1,21 @@
...
+
+    service: str = None
+    target: str = None
+    config: Dict = None
+    addrs: List[str] = None
+    priority: int = None
+    topic: str = None
+    title: str = None
+    message: Union[str, bytes] = None
+    data: Dict = None
```
amotl, line 21: Future versions of mqttwarn will also rely on instances of this class instead of the dictionary which is currently used to propagate information through the message queue to the worker threads.

jpmens (approved):

PR #538, opened 2021-06-19
Add publishversion - new mqtt topic upon mqttwarn start up

psyciknz:
Upon starting, and with the configuration topic publishversion with an mqtt topic, mqttwarn will publish it's current version to the set topic.

Eg:
```
david@melkor:~$  $ mosquitto_sub  -v -t mqttwarn/version
mqttwarn/version 0.26.1
```

- Updated the Handbook.md to describe the setting
- Added a try catch around the cofiguration item cf.publishversion not existing (will just skip)

Due to my runtime issues #516 I did try adding a test for connection....you'll see it started by commented out, I hope this is a prompt to complete this test as I am not experienced enough to complete it.  Ideally there would be tests with a connection success and failure (to servers) and this could then test other mqtt parameters (passwords, lwt and the publishing of publish version).


jpmens:
Can you tell us a bit about the motivation for the feature? What programs (or people?) would benefit from mqttwarn publishing version information at startup? It's not really as though any program could actually react to a different number there, or could it?

psyciknz:
No but it's handy reference information, especially if you're an automated docker shop with watchdog etc renewing your containers.

amotl:
Dear @psyciknz,

thank you for bringing up this idea. I cherry-picked your commit 2d96ff8, converged it into a more generic implementation, and submitted it through #552. I hope you like it.

With kind regards,
Andreas.

PR #542, opened 2021-09-29
fix xbmc/kodi service when authentication enabled (on Python 3)

### LINKED ISSUE #541: XBMC/Kodi service broken if authentication enabled (on Python 3)
Getting the following error;
```
ERROR [xbmc] Error sending XBMC notification to kodi [192.168.10.40:8080]: module 'base64' has no attribute 'encodestring'
```

It appears that `base64.encodestring()` has been deprecated;

> In Python 3 encodestring docs says:
>
>def encodestring(s): """Legacy alias of encodebytes().""" import warnings warnings.warn("encodestring() is a deprecated alias, use encodebytes()", DeprecationWarning, 2) return encodebytes(s)

According to this [link](https://stackoverflow.com/questions/23164058/how-to-encode-text-to-base64-in-python) we should be doing something like;
```
        if xbmcpassword is not None:
            credentials = '%s:%s' % (xbmcusername, xbmcpassword)
            base64string = base64.b64encode(credentials.encode('utf-8')).decode('utf-8')
            authheader = "Basic %s" % base64string
            req.add_header("Authorization", authheader)
```

sumnerboy12:
closes #541

### DIFF
--- a/mqttwarn/services/xbmc.py
+++ b/mqttwarn/services/xbmc.py
@@ -51,7 +51,8 @@ def plugin(srv, item):
         req = urllib.request.Request(url, jsoncommand)
         req.add_header("Content-type", "application/json")
         if xbmcpassword is not None:
-            base64string = base64.encodestring('%s:%s' % (xbmcusername, xbmcpassword))[:-1]
+            credentials = '%s:%s' % (xbmcusername, xbmcpassword)
+            base64string = base64.b64encode(credentials.encode('utf-8')).decode('utf-8')
             authheader = "Basic %s" % base64string
             req.add_header("Authorization", authheader)
         response = urllib.request.urlopen(req, timeout = 2)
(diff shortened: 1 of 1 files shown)


mqttwarn/services/xbmc.py
```diff
@@ -51,7 +51,8 @@ def plugin(srv, item):
         req = urllib.request.Request(url, jsoncommand)
         req.add_header("Content-type", "application/json")
         if xbmcpassword is not None:
-            base64string = base64.encodestring('%s:%s' % (xbmcusername, xbmcpassword))[:-1]
+            credentials = '%s:%s' % (xbmcusername, xbmcpassword)
+            base64string = base64.b64encode(credentials.encode('utf-8')).decode('utf-8')
```
amotl, line 55: No need to apply decoding from utf-8 here, just cast to bytes.
```suggestion
            base64string = base64.b64encode(credentials.encode('utf-8')).decode()
```

amotl:
Dear Ben,

based on your patch, I expanded your fix to the service adapters for HTTP and Ionic, see #543. I also just published a new bugfix release mqttwarn 0.26.2, including your fixes.

With kind regards,
Andreas.

sumnerboy12:
Great thanks!

PR #543, opened 2021-09-29
Fix HTTP basic authentication for Python 3

amotl:
Hi there,

@sumnerboy12 already contributed #542 in order to fix an issue with missing compatibility for Python 3. This patch applies his fix to other plugins as well.

With kind regards,
Andreas.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,13 @@ in progress
 ===========
 
 
+2021-09-29 0.26.2
+=================
+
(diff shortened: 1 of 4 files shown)


sumnerboy12:
Hi @amotl thanks for porting this fix to the other services - sorry I should have checked!

Could you trigger a new Docker image for 0.26.2 please? I need to update my local image so my XBMC notifications start working again :).

jpmens:
@sumnerboy12 I thought I could just press a button, but that doesn't work unfortunately. I'm also far away from everywhere, so we'll have to wait for @amotl to punch the knobs.

Issue #544, opened 2021-09-30
Clarify which Docker images are regularily published


amotl:
> Hi @amotl,
> 
> could you trigger a new Docker image for 0.26.2 please? I need to update my local image so my XBMC notifications start working again :).

_Originally posted by @sumnerboy12 in https://github.com/jpmens/mqttwarn/issues/543#issuecomment-930501754_

amotl:
@jpmens answered:

> @sumnerboy12 I thought I could just press a button, but that doesn't work unfortunately. I'm also far away from everywhere, so we'll have to wait for @amotl to punch the knobs.

amotl:
Hi there,

after some improvements recently, CI will automatically push Docker images to GitHub Container Registry (GHCR) [1]. I don't actually know the state about the images on Docker Hub. We can also automatically push them to Docker Hub on behalf of [2,3] after discussing with @jpmens how to authenticate there.

Also, we might want to update the documentation about how to consume Docker images appropriately in the README file of the project. Until then, please consult the Docker handbook [4].

With kind regards,
Andreas.

[1] https://github.com/jpmens?tab=packages&repo_name=mqttwarn
[2] https://github.com/jpmens/mqttwarn/blob/main/.github/workflows/release-docker-standard.yml
[3] https://github.com/jpmens/mqttwarn/blob/main/.github/workflows/release-docker-full.yml
[4] https://github.com/jpmens/mqttwarn/blob/main/DOCKER.md

sumnerboy12:
I would imagine most users will be using DockerHub - if not all? Would it be a lot of work to add this to CI?

amotl:
Sure, we can do it. I will have a conversation with @jpmens about it.

sumnerboy12:
Great - thanks Andreas.

amotl:
I believe it should just work by using this alternative image name. It should be no further configuration required.

On 11 October 2021 11:19:09 CEST, Ben Jones ***@***.***> wrote:
>I am running mqttwarn in a docker swarm, deployed by ansible - I am not
>sure how to use GHCR but will do some reading.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/jpmens/mqttwarn/issues/544#issuecomment-939845748

-- 
Sent from my mind. This might have been typed on a mobile device, so please excuse my brevity.

sumnerboy12:
Thanks @amotl - that worked!

[...]

PR #551, opened 2021-10-17
Update documentation

amotl:
This patch is mostly about clarifying which is currently the recommended Docker image, until automatic publishing to Docker Hub will be implemented. See also #544.

/cc @jpmens, @sumnerboy12

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -9,22 +9,22 @@ run it interactively, perhaps to help diagnose a problem.
 
 Docker images are automatically published to:
 
-- https://hub.docker.com/r/jpmens/mqttwarn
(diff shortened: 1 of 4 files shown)


sumnerboy12:
Looks good @amotl. Could I suggest you update the `Dockerfile.mqttwarn-slack` example to use `FROM ghcr.io/jpmens/mqttwarn-standard:latest`? 

Otherwise this example very quickly gets out of date (it is currently pointing at 0.27.0).

amotl:
Dear Ben,

I've just submitted b707341cb to implement your suggestion.

With kind regards,
Andreas.

PR #572, opened 2022-08-18
Chore: Improve exception handling

amotl:
What the title says. This patch should not change any significant behavior of mqttwarn, it just improves the logging a bit.

Reference: #571.


jpmens (approved):
LGTM

PR #576, opened 2022-08-21
Add test for exercising configuration file scaffolding

amotl:
This test exercises the problem on Windows10 reported by @symonjim at #571.


amotl:
Apparently, when using Bash, the problem does not occur.

https://github.com/jpmens/mqttwarn/blob/bfd177b6cc175a15abfb51eff1dab8d963081a93/.github/workflows/tests.yml#L18-L20

amotl:
Apparently, it also works on GHA/Win10/PowerShell. Weird.

-- https://github.com/jpmens/mqttwarn/runs/7940917310

PR #581, opened 2022-08-29
Complete rewrite of osxnotify

portalzine:
Use of desktop-notifier https://github.com/samschott/desktop-notifier
Currently synchronous, with limited options. 
Moving to an asynchronous notification, would open up callbacks, buttons and more.

PR #586, opened 2022-09-15
Add tests for `osxnotify` plugin, fix evaluating `sound` config setting

amotl:
This patch adds a few test cases to exercise the `osxnotify` plugin after the rewrite #581 by @portalzine. While being at it, it fixes an evaluation glitch with the `sound` configuration setting. `config == dict` would never evaluate to `True`.


jpmens (approved):
LGTM

PR #588, opened 2022-09-15
Improve documentation and `setup.py` for `osxnotify` service plugin

amotl:
A small patch to improve the documentation and setup convenience around the `osxnotify` rework.

/cc @portalzine

### DIFF
--- a/HANDBOOK.md
+++ b/HANDBOOK.md
@@ -2004,7 +2004,8 @@ Requires:
 
 ### `osxnotify`
 
-* For Macs only, utilizing desktop-notifier from https://github.com/samschott/desktop-notifier 
+* Despite its name, it is a cross-platform desktop notifier, based on
+  https://github.com/samschott/desktop-notifier.
 
 ```ini
 [config:osxnotify]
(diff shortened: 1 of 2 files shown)


jpmens (approved):

portalzine:
Was thinking about proposing another documentation update as well, but wanted to stick with the original. Might be worth moving the plugin to a new name in the future, like "desktopnotify". Thanks for the update.

jpmens:
I second the name change to `desktopnotify` in order to remove the inaccurate operating system denomination from the current name.

amotl:
I totally agree with the name change. However, I am proposing to call the service just `desktop.py`. I think the "notify" suffix is somehow redundant in this context. Let me know what you think about it. In any case, I can submit a corresponding patch.

portalzine:
I agree,  "..notify" is a bit redundant, so desktop.py works 👍

jpmens:
I feel `desktop` is too generic as we might have other notifiers for desktops. :)

portalzine:
toast , bubble or popup_notifier

amotl:
> I second the name change to `desktopnotify` [...] and feel `desktop` is too generic.

I am also fine with that decision. Let's just do it like that.

PR #609, opened 2022-11-12
Make the PushOver API retry and expire configurable.

jlrgraham:
Add support for setting the `retry` and `expire` elements in the PushOver API: https://pushover.net/api


jpmens:
Merged, thank you!

PR #608, opened 2022-11-12
Move filter skipped messages to debug level, reduce chatter on busy f…

jlrgraham:
…ilters.

I'd propose decreasing the severity of this log message.  I've got a filter setup that is skipping messages on a relatively busy broker and I'm winding up with logs that are mainly filter skip messages.

I'd also considered trying to make this a configurable option; maybe adding a `config.skipped_messages_log_level` setting, but wanted to propose here first.  If that's preferred please let me know and I'd update the PR here.  Thanks!

PR #611, opened 2022-11-20
Note filteredmessagesloglevel in example config.

jlrgraham:
Fixing from comments in: https://github.com/jpmens/mqttwarn/pull/608#issuecomment-1321139389


amotl:
Thank you!

Issue #628, opened 2023-03-12
Proposal: Improve configuration for service targets


amotl:
Dear JP and Ben,

while modernizing mqttwarn on some aspects after taking over maintenance, I regularily touch the aspect about the configuration of the _plugin service targets_ details, currently being a `list`, mostly. I believe it may be confusing for service plugins which would better need to obtain _named parameters_ instead. I would like to make a general proposal, in order to improve this situation.

With kind regards,
Andreas.

### Status quo

For the SMTP service, this is just perfect.
```ini
[config:smtp]
targets = {
    'localj'     : [ 'jpm@localhost' ],
    'special'    : [ 'ben@gmail', 'suzie@example.net' ]
    }
```

At the Pushover configuration snippet, for example, it is tolerable.
```ini
[config:pushover]
targets = {
    'nagios'     : ['userkey1', 'appkey1', 'sound1'],
    'alerts'     : ['userkey2', 'appkey2'],
    'tracking'   : ['userkey1', 'appkey2', None, 'cellphone1,cellphone2'],
    'extraphone' : ['userkey2', 'appkey3']
    }
```

On the other hand, it is definitively confusing for the Pushsafer plugin, where you would have to constantly maintain a mind map of parameter positions within the list.
```ini
[config:pushsafer]
targets = {
    'nagios'     : ['privatekey', 'Device ID', 'Icon', 'Sound', 'Vibration', 'URL', 'Url Title', 'Time2Live', 'Priority', 'Retry', 'Expire', 'Answer'],
    'tracking'   : ['aliaskey1'],
    'extraphone' : ['aliaskey2', '', '', '', '', '', '', '60', '2', '60', '600', '0'],
    'warnme'     : ['aliaskey3', '', '', '', '', '', '', '60', '1', '', '', '1']
    }
```

### Proposal

So, my proposal is to use dictionaries instead, **where it makes sense** to propagate named parameters to service plugins, similarly to how it has been implemented for the Apprise service plugin already, see GH-558 and GH-627.

```ini
[config:apprise]
; Dispatch message to multiple Apprise plugins.
module   = 'apprise_multi'
targets = {
   'demo-mailto'      : [ {
          'baseuri':  'mailtos://smtp_username:smtp_password@mail.example.org',
          'recipients': ['foo@example.org', 'bar@example.org'],
          'sender': 'monitoring@example.org',
          'sender_name': 'Example Monitoring',
          } ],
   }
```

[...]

PR #629, opened 2023-03-12
[Pushsafer] Improve configuration to use named parameters

amotl:
Hi,

following the rationale provided at GH-628, this would be my proposal to improve the Pushsafer configuration snippet to use named parameters instead of a list, which is confusing, and barely maintainable and extendable.

### Documentation
Visit the rendered version of the reworked [mqttwarn » Pushsafer service](https://github.com/jpmens/mqttwarn/blob/d9c9efb32df75e31dea3c10fd68c6de1122143b0/HANDBOOK.md#pushsafer).

### Configuration snippet
A valid configuration for Pushsafer in the v2 layout would be:
```ini
[config:pushsafer]
; https://www.pushsafer.com/en/pushapi
; https://www.pushsafer.com/en/pushapi_ext
targets = {
    'basic': { 'private_key': '3SAz1a2iTYsh19eXIMiO' },
    'nagios': {
        'private_key': '3SAz1a2iTYsh19eXIMiO',
        'device': '52|65|78',
        'icon': 64,
        'sound': 2,
        'vibration': 1,
        'url': 'http://example.org',
        'url_title': 'Example Org',
        'time_to_live': 60,
        'priority': 2,
        'retry': 60,
        'expire': 600,
        'answer': 1,
        },
    'tracking': {
        'private_key': '3SAz1a2iTYsh19eXIMiO',
        'device': 'gs23',
        'icon': 18,
        },
    'bogus': { 'foo': 'bar' },
    }
```

With kind regards,
Andreas.

/cc @appzer

### DIFF
--- a/mqttwarn/services/pushsafer.py
+++ b/mqttwarn/services/pushsafer.py
@@ -36,6 +36,9 @@ def pushsafer(**kwargs):
     if not kwargs['k']:
         kwargs['k'] = os.environ['PUSHSAFER_TOKEN']
 
+    # Don't submit empty parameters to Pushsafer.
+    filter_empty_parameters(kwargs)
+
     url = urllib.parse.urljoin(PUSHSAFER_API, "api")
     data = urllib.parse.urlencode(kwargs).encode('utf-8')
     req = urllib.request.Request(url, data)
@@ -45,6 +48,16 @@ def pushsafer(**kwargs):
 
     if data['status'] != 1:
         raise PushsaferError(output)
+
+
+def filter_empty_parameters(params: t.Dict[str, str]):
+    """
+    Filter empty parameters.
+    """
+    for key in list(params.keys()):
+        value = params[key]
+        if value == "":
+            del params[key]
 
 
 def plugin(srv, item):
@@ -132,7 +145,12 @@ class PushsaferParameterEncoder:
             params['s'] = addrs[3]
 
         if len(addrs) > 4:
-            params['v'] = addrs[4]
+            # Special handling for `vibration`.
+            # With the v1 configuration layout, it should not always trigger a vibration when left empty.
+            # Therefore, a synthetic value `0` can be used to configure "device-default" vibration.
+            value = str(addrs[4])
+            if value != "":
+                params['v'] = value
 
         if len(addrs) > 5:
             params['u'] = addrs[5]
--- a/mqttwarn/services/pushsafer.py
+++ b/mqttwarn/services/pushsafer.py
@@ -5,6 +5,9 @@ __author__    = 'Jan-Piet Mens <jpmens()gmail.com>'
 __copyright__ = 'Copyright 2014 Jan-Piet Mens'
 __license__   = 'Eclipse Public License - v 1.0 (http://www.eclipse.org/legal/epl-v10.html)'
 
+import dataclasses
+from collections import OrderedDict
+
 from future import standard_library
 standard_library.install_aliases()
 import os
@@ -83,7 +86,7 @@ class PushsaferParameterEncoder:
         if isinstance(addrs, t.List):
             self.encode_v1()
         elif isinstance(addrs, t.Dict):
-            raise NotImplementedError("Pushsafer configuration layout v2 not implemented yet")
+            self.encode_v2()
         else:
             raise ValueError(f"Unable to decode Pushsafer configuration layout. type={type(addrs)}")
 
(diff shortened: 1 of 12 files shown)


tests/etc/better-addresses.ini
```diff
@@ -67,6 +67,33 @@ targets = {
...
+[config:pushsafer]
+; https://www.pushsafer.com/en/pushapi
+; https://www.pushsafer.com/en/pushapi_ext
+targets = {
+    'nagios': {
+        'private_key': '3SAz1a2iTYsh19eXIMiO',
+        'device': '52|65|78',
+        'icon': 64,
+        'sound': 2,
+        'vibration': 'blank',  # ?
```
amotl, line 79: Dear @appzer,

may I humbly ask you to review this line?
```
'vibration': 'blank',  # ?
```

The documentation [^1] says about the value range:
> `blank`=device default or a number `1`-`3`.

Does it actually mean that the client should send `&v=` in order to enable the vibration feature with device defaults, without any value?

With kind regards,
Andreas.

[^1]: https://www.pushsafer.com/en/pushapi_ext#API-V

appzer: > &v=

that correct! BLANK means without value = device default!

amotl: Thank you. I will take care about proper serialization of that parameter, which will have to be handled specially, because `None` values would otherwise not be propagated.

May I ask if there are **any other parameters like this**, where including them with an empty value into the query string actually signals something?

amotl: b021547a75d3 now implements the special handling for the `vibration` parameter, _to make sure it is submitted empty_, while all the other parameters will be omitted when empty.

In order to use the device-default vibration, mqttwarn users will have to explicitly configure `vibration=0` now. I think this slight breaking change is acceptable while fixing an actual bug, that previously, vibrations have been turned on for each notification.

-- https://www.pushsafer.com/en/pushapi_ext#API-V

mqttwarn/services/pushsafer.py
```diff
@@ -36,82 +50,217 @@ def pushsafer(**kwargs):
...
+
+        # Decode Private or Alias Key.
+        try:
+            self.private_key = addrs["private_key"]
+        except KeyError:
+            raise PushsaferConfigurationError(f"Pushsafer private or alias key not configured")
+
+        params = {
+            'expire': 3600,
+        }
```
amotl, line 205: Just out of curiosity: Do you think it is still fine to set a default expiration value of 3600 seconds? It might be unexpected to the user. On the other hand, it might have had a reason to implement it like that. Can you remember that, @jpmens or @sumnerboy12?

-- https://www.pushsafer.com/en/pushapi_ext#API-EX

amotl: > It might be unexpected to the user.

I've documented the current default value with 91881285efde0de2.

sumnerboy12: I have no recollection of why that default was chosen. It does seem very long tho.

amotl: Thanks for your reply, Ben. The documentation currently does not tell anything about a default value. Would omitting that parameter actually mean that notifications would never expire? Maybe this was the exact reason why that default was chosen -- to have at least some kind of limit?

However, again depending on the use case and scenario, I think 24h may also be a reasonable default value. What if you would be on a flight, for let's say eight hours, or have your phone turned off during the night? Would you completely miss the notifications because they would expire before having the chance to actually get delivered?

What do you think? Maybe removing that default value would make more sense to cover those situations, if you agree that 3600 seconds may actually well be too short instead.

amotl: > Maybe removing that default value would make more sense to cover those situations, if you agree that 3600 seconds may actually well be too short instead.

In this case, mqttwarn would not interfere at all, and let the user choose the intended value independently.

tests/etc/better-addresses.ini
appzer: Hi, 

yes, the parameter sound can be empty for device default sound.
0 = silent
1-62 plays a pushsafer sound

I cant change the API handling.

Kind Regards Kevin

mqttwarn/services/pushsafer.py
appzer: expire=3600 makes no sense that the expire value is passed here

```
if len(addrs) > 10:
            params['ex'] = addrs[10]
```

tests/etc/better-addresses.ini
appzer: you also can use the special handling for the sound parameter!

Thanks Kevin

amotl: > The parameter `sound` can be empty for device-default sound.

I see. Thanks for the clarification.

> I cant change the API handling.

I understand.

mqttwarn/services/pushsafer.py
amotl: @appzer: 3600 is currently the default value, it can be overwritten when using the `expiration` parameter.

appzer: but the expire parameter also needs the resend parameter to work
https://www.pushsafer.com/en/pushapi_ext#API-EX

amotl: Oh I see, thank you so much for telling me how the API works. I actually never used Pushsafer, apologies. In this case, it may make even more sense to remove that ambiguous default value, if no objections will be raised. /cc @jpmens

amotl: a40efe52ae removes this default value. Thank you again.

tests/etc/better-addresses.ini
amotl: Regarding the _Sound_ parameter, the documentation says:
> If the parameter is not passed or left blank, the device plays the default sound.

I think in this case it is fine, because the default behavior actually is to play the default sound _when the parameter is omitted_.

Maybe we have a misunderstanding here: Would it actually have been fine to omit the `vibration` parameter as well?

-- https://www.pushsafer.com/en/pushapi_ext#API-S

amotl: Maybe we have a misunderstanding here: Would it actually have been fine to omit the vibration parameter as well? Oh well, I am right now re-reading the documentation, and it clearly states:

> **How often** the device should vibrate when receiving a push-notification.

I overlooked that detail. Apologies. I guess I will revert all my changes about this then. I misinterpreted the parameter to be a able to turn vibrations off.

-- https://www.pushsafer.com/en/pushapi_ext#API-V

amotl: > Does it actually mean that the client should send `&v=` in order to enable the vibration feature with device defaults, without any value?

This was my initial question. Can you confirm it does not make any difference if the client sends `&v=`, or just omit it?

amotl: 6e68f7faa removes the special handling again, if you agree. Sorry again for the noise.

appzer: > This was my initial question. Can you confirm it does not make any difference if the client sends `&v=`, or just omit it?

that makes no difference

appzer:
@amotl  do you also add the 3 new parameter (cr, ao, af) to your rework? Thanks Kevin

amotl:
> Do you also add the 3 new parameter (cr, ao, af) to your rework?

They will be added separately, based on the patch you've submitted. Thanks.

amotl:
Hi again. GH-630 has been adjusted accordingly. Let's see what @jpmens and @sumnerboy12 will have to say about both patches.

sumnerboy12 (approved):
Looks to be a significant improvement.

amotl:
Hi again,

I've just released mqttwarn 0.33.0 [^1], including all corresponding improvements around Pushsafer. Apologies for the delay, and thanks again for your contributions.

With kind regards,
Andreas.

[^1]: https://github.com/jpmens/mqttwarn/releases/tag/0.33.0

Issue #634, opened 2023-04-12
Non-UTF-8 encoding causes error


sevmonster:
Currently mqttwarn expects a message encoded with UTF-8. Could it be possible to specify the encoding? I am trying to save a JPEG being sent over MQTT, and since it is obciously not UTF-8 encoded, mqttwarn exits with an error.

## `mqttwarn.ini`
```toml
[config:file]
append_newline = False
overwrite = True
targets = {                                                              'cam1-person': ['/media/snapshots/cam1-person.jpg'],
      'cam2-person': ['/media/snapshots/cam2-person.jpg']
    }
                                                                   [frigate/cam1/person/snapshot]
targets = file:cam1-person
[frigate/cam2/person/snapshot]
targets = file:cam2-person
```

## Error
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
Traceback (most recent call last):                                   File "/usr/local/bin/mqttwarn", line 8, in <module>
    sys.exit(run())
             ^^^^^
  File "/usr/local/lib/python3.11/site-packages/mqttwarn/commands.py", line 91, in run
    run_mqttwarn()
  File "/usr/local/lib/python3.11/site-packages/mqttwarn/commands.py", line 147, in run_mqttwarn                                          subscribe_forever()
  File "/usr/local/lib/python3.11/site-packages/mqttwarn/core.py", line 643, in subscribe_forever
    mqttc.loop_forever()
  File "/usr/local/lib/python3.11/site-packages/paho/mqtt/client.py", line 1756, in loop_forever
    rc = self._loop(timeout)
         ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/paho/mqtt/client.py", line 1164, in _loop
    rc = self.loop_read()                                                   ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/paho/mqtt/client.py", line 1556, in loop_read                                             rc = self._packet_read()
         ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/paho/mqtt/client.py", line 2439, in _packet_read                                          rc = self._packet_handle()
         ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/paho/mqtt/client.py", line 3033, in _packet_handle                                        return self._handle_publish()
[...]

Issue #632, opened 2023-04-06
Help with my configuration?


sevmonster:
I set up a Frigate -> Mosquitto -> mqttwarn -> Apprise -> Ntfy -> Nginx flow (whew) and everything is working... Almost. The filter seems to not be blocking bad messages even though it should be based on my debugging. And, I can't figure out how to send attachments. My only lead is to save the attachment to a local file before using the `attach` query parameter in my Ntfy URI, but it would be nice to be able to send it through the body of the message instead, which I believe Apprise supports.

Depending on if any of this not working is a bug or potential feature request, this issue could turn into an actual issue and not a support request :)

# `mqttwarn.ini`
```ini
[defaults]
functions = funcs.py
launch    = apprise-ntfy

status_publish = True


[config:apprise-ntfy]
module   = apprise_single
baseuri  = ntfys://ntfy/frigate


[frigate/events]
filter  = frigate_events_filter()
alldata = frigate_events()
targets = apprise-ntfy
title   = {title}
format  = {format}
click   = {click}
```

# `funcs.py`
```python
# -*- coding: utf-8 -*-
from datetime import datetime

try:
    import json
except ImportError:
    import simplejson as json

def frigate_events(topic, data, srv=None):
    if srv is not None:
        srv.logging.debug('events ******************************************')
    a = json.loads(data['payload'])['after']
    f = lambda x: (y.replace('_', ' ') for y in x)
    r = {
      'camera': a['camera'],
      'label':  a['sub_label'] or a['label'],
      'current_zones': ', '.join(f(a['current_zones'])),
      'entered_zones': ', '.join(f(a['entered_zones'])),
      'time':   datetime.fromtimestamp(a['frame_time'])
    }
    r.update({
      'title':  f"{r['label']} entered {r['entered_zones']}",
      'format': f"In zones {r['current_zones']} at {r['time']}",
      'click':  f"https://frigate/events?camera={r['camera']}&label={r['label']}&zone={a['entered_zones'][0]}"
    })
    return r

def frigate_events_filter(topic, message, section, srv=None):
    if srv is not None:
        srv.logging.debug('filter ******************************************')
    try:
        message = json.loads(message)
    finally:
        message = None
    if message and message.get('type', None) != 'end' and 'after' in message:
        a = message['after']
        return False not in (x in a and (x == 'current_zones' or a[x]) for x in
                                  ('false_positive', 'camera', 'label',
                                   'current_zones', 'entered_zones',
[...]

PR #635, opened 2023-04-12
Forwarding Frigate events to Ntfy

amotl:
### About
@sevmonster came up with an interesting use-case, aiming to subscribe to events from [Frigate](https://frigate.video/), and forward them to Apprise/Ntfy.

### What's inside
This patch brings in the configuration and adapter files into the `examples` directory (thanks, @sevmonster!), and may also include corresponding commits to unlock one thing or another.

### References
- GH-632
- GH-634


examples/frigate/frigate.py
```diff
@@ -0,0 +1,41 @@
...
+from datetime import datetime
+
+try:
+    import json
+except ImportError:
+    import simplejson as json
+
+def frigate_events(topic, data, srv=None):
+    if srv is not None:
+        srv.logging.debug('events ******************************************')
```
sevmonster, line 11: You can probably remove these debug messages.

examples/frigate/frigate.py
```diff
@@ -0,0 +1,41 @@
...
+    r.update({
+      'title':  f"{r['label']} entered {r['entered_zones']}",
+      'format': f"In zones {r['current_zones']} at {r['time']}",
+      'click':  f"https://frigate/events?camera={r['camera']}&label={r['label']}&zone={a['entered_zones'][0]}"
+    })
+    return r
+
+def frigate_events_filter(topic, message, section, srv=None):
+    if srv is not None:
+        srv.logging.debug('filter ******************************************')
```
sevmonster, line 30: You can probably remove these debug messages.

examples/frigate/frigate.py
amotl: Addressed with c2623646b, thanks.

examples/frigate/frigate.py
amotl: Addressed with c2623646b, thanks.

examples/frigate/frigate.py
```diff
@@ -23,15 +25,43 @@ def frigate_events(topic, data, srv=None):
...
-    if message and message.get('type', None) != 'end' and 'after' in message:
-        a = message['after']
+        message = dict()
+    message_type = message.get("type", None)
+    srv.logging.info(f"Received Frigate event message with type={message_type}")
+
+    # Determine if message should be used, or not.
+    use_message = message_type == "new" and "after" in message
+
+    # Look at more details.
```
sevmonster, line 48: These changes are likely much better for an example, rather than my arcane list comprehensions :)

Looks good

amotl: Thanks. As outlined over at https://github.com/jpmens/mqttwarn/issues/632#issuecomment-1506010580, 58f716d58 attempts to bring your most recent snippet into this shape. I hope it will be for the better.

examples/frigate/frigate.py
```diff
@@ -0,0 +1,143 @@
...
+        format=f"In zones {event.current_zones_str} at {event.time}",
+        click=f"https://frigate/events?camera={event.camera}&label={event.label}&zone={event.entered_zones[0]}",
+        attach=attach_filename,
+    )
+    return ntfy_parameters.to_dict()
+
+
+def frigate_events_filter(topic, message, section, srv: Service):
+    """
+    mqttwarn filter function to only use Frigate events of type `new`.
```
sevmonster, line 92: Technically this handles both `new` and `update`. There are 3 `type`s, of `new`, `update`, and `end`, so the filter will handle both `new` and `update` if it only excludes `end`.

examples/frigate/frigate.py
```diff
@@ -0,0 +1,143 @@
...
+                srv.logging.warning("Frigate event skipped, it is a false positive")
+                return True
+            else:
+                continue
+
+        # All other keys should be present and have values.
+        if not value:
+            srv.logging.warning(f"Frigate event skipped, field is empty: {field}")
+            return True
+
```
sevmonster, line 142: End users (such as myself) may also want to further filter `update` messages by checking if `current_zones` is equal to `entered_zones`; this can happen if the snapshot is updated or its state has changed, but the object has not moved to a new zone. For example, you might add something like this:
```python
if message['type'] == 'update' and after['current_zones'] == after['entered_zones']:
    srv.logging.warning("Frigate event skipped: Object didn't enter new zone")
```

mqttwarn/context.py
sevmonster, line 1: I like the increase in priority for the logs. I would certainly consider them more critical than warnings.

sevmonster (changes requested):

examples/frigate/frigate.py
```diff
@@ -0,0 +1,143 @@
...
+        entered_zones=after['entered_zones'],
+    )
+
+    # Interpolate event data into attachment filename template.
+    attach_filename = filename_template.format(**event.to_dict())
+
+    # Compute parameters for outbound Apprise / Ntfy URL.
+    ntfy_parameters = NtfyParameters(
+        title=f"{event.label} entered {event.entered_zones_str}",
+        format=f"In zones {event.current_zones_str} at {event.time}",
```
sevmonster, line 83: I revised this in my own configuration to make it read better when `{event.current_zones_str}` is null. But the current one may work better for example as it is simpler.
```suggestion
        format=f"At {event.time}{', was in ' + event.current_zones_str if event.current_zones else ''}",
```

amotl:
> We have success. I had to make the changes I suggested in the PR to get the attachment to go through and prevent repeated notifications. But otherwise this is great.
>
> -- https://github.com/jpmens/mqttwarn/issues/632#issuecomment-1507841463

Excellent! Thanks for adding your suggestions, I will try to bring them in. May I ask you to also share your walkthrough? It may be useful to spice up the README, specifically if you are using a setup based on Docker, which I did not take into consideration yet. And of course, this is the area where things may still want to be improved, with respect to your comments about addressing the filename/URL of the attachment.

examples/frigate/publish.sh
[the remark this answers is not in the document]
amotl: Can you also change the code to use the `filename` parameter instead of `attach`, and try a few iterations with that?

-- https://github.com/jpmens/mqttwarn/issues/632#issuecomment-1504199365

examples/frigate/frigate.py
```diff
@@ -0,0 +1,143 @@
...
+        click=f"https://frigate/events?camera={event.camera}&label={event.label}&zone={event.entered_zones[0]}",
+        attach=attach_filename,
+    )
+    return ntfy_parameters.to_dict()
```
amotl, line 87: In order to quickly patch propagating the corresponding file reference to Apprise/Ntfy as `filename` instead of `attach`.

```suggestion
    params = ntfy_parameters.to_dict()
    params["filename"] = params["attach"]
    del params["attach"]
    return params
```

examples/frigate/frigate.ini
```diff
@@ -0,0 +1,56 @@
...
+
+# Format: JSON
+# Docs:   https://docs.frigate.video/integrations/mqtt/#frigateevents
+
+
+
+[config:apprise-ntfy]
+module   = apprise_single
+baseuri  = ntfys://ntfy/frigate
+filename_template = './var/media/{camera}-{label}.jpg'
```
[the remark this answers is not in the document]
amotl: See also my comment above, at https://github.com/jpmens/mqttwarn/pull/635#discussion_r1166670893. This would use the [another code path](https://github.com/binwiederhier/ntfy/blob/1ccf6597815b7b3bd7a9b7f30a6d6d60c4620798/server/server.go#L826-L828) instead.

amotl:
I've just familiarized myself a bit with Ntfy, and found that example how to attach a local file [^1]. It looks like Ntfy has all the machinery in place to accept and store files, and provide them again via HTTP, so this is indeed what should be leveraged here.

Is it only the current Ntfy implementation based on Apprise, which is not permitting to use the feature like that?

[^1]: https://docs.ntfy.sh/subscribe/cli/#attaching-a-local-file

amotl:
I've not further explored this space, but I quickly want to reference some pointers. As far as I can see, the Apprise/Ntfy code currently implements the HTTP headers `X-{Icon,Priority,Delay,Click,Email,Tags}`. However, the Ntfy client code, probably used for `ntfy pub`, also implements a few others, like `X-{Title,Actions,Attach,Filename,Cache,Firebase}`.

So, could using `X-Attach` or `X-Filename` unlock anything sensible here, which would improve the situation?

[^1]: https://github.com/caronc/apprise/blob/v1.3.0/apprise/plugins/NotifyNtfy.py#L502-L542
[^2]: https://github.com/binwiederhier/ntfy/blob/main/client/options.go#L20-L103

amotl:
I've discovered the Python client implementation [ntfy-wrapper](https://github.com/vict0rsch/ntfy-wrapper). This is the corresponding code which implements the "attach" path. TLDR; It uses the HTTP PUT method in this case.

- https://github.com/vict0rsch/ntfy-wrapper/blob/8d5e28c0ba747a660e058e43954e807eeccb1450/ntfy_wrapper/notifier.py#L355-L356
- https://github.com/vict0rsch/ntfy-wrapper/blob/8d5e28c0ba747a660e058e43954e807eeccb1450/ntfy_wrapper/notifier.py#L373-L377
- https://github.com/vict0rsch/ntfy-wrapper/blob/8d5e28c0ba747a660e058e43954e807eeccb1450/ntfy_wrapper/notifier.py#L426-L431

examples/frigate/publish.sh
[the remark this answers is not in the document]
amotl: Please consider this comment as obsolete. I've not been able to discover any code in Apprise/Ntfy, which actually uploads the attachment. So, no matter how it will be stuffed with parameters, the code will probably never do what we are aiming it. See also discussion at https://github.com/jpmens/mqttwarn/pull/635#issuecomment-1508391174 ff.

examples/frigate/frigate.ini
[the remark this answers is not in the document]
amotl: Please consider this comment as obsolete, see https://github.com/jpmens/mqttwarn/pull/635#discussion_r1166789969.

mqttwarn/context.py
amotl: Yeah, and the error output will even have tracebacks now.

examples/frigate/frigate.py
amotl: Thanks. Feel free to add any improvements on this detail. I am not so much into those aspects right now.

examples/frigate/frigate.py
```diff
@@ -0,0 +1,143 @@
...
+    :return: True if message should be filtered, i.e. notification should be skipped.
+    """
+    try:
+        message = json.loads(message)
+    except json.JSONDecodeError as e:
+        srv.logging.warning(f"Can't parse Frigate event message: {e}")
+        return True
+
+    # ignore ending messages
+    if message.get('type', None) == 'end':
```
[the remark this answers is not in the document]
amotl: > You have also reminded me that a second argument of None to dict.get is not required since it defaults to None. (Has it always been like that? I thought it raised KeyError.)

Yeah, getting an item out of a dictionary with `.get()` always returned `None` I guess, when the key was not available.

The `KeyError` will happen if you access the outcome as a dictionary. So, the only reasonable thing in this case would be to use it like `message.get('<key>', {})`, if you want to write your code in a graceful manner, when aiming to continue working on a dictionary type, as if the item would have been present.

examples/frigate/frigate.py
sevmonster:
```suggestion
    mqttwarn filter function to only use new events and important updates.
```

sevmonster:
> Please let me know if you see any problems with this approach.

There aren't any problems with your approach necessarily, but there are potential blockers elsewhere. To reiterate/condense these:
- Frigate does not send `thumbnail` through the MQTT event, so the file will have to be loaded from somewhere elae if it is to be send along with the message
- Apprise as far as I can tell does not support passing body to Ntfy, and only the `attachment` parameter

I initially suggested sending the attachment through the request body in the OP of this issue, before I realized Apprise most likely didn't support it.

I see three options to improve the situation:
- Pushing code/issues to the above blocking projects to fix the stated issues
- Bundle the `ntfy-wrapper` package or similar into `mqtt-full` and create a new mqttwarn service to handle it
- Use `requests` or similar and just write out the raw Ntfy HTTP API request into a mqttwarn service

amotl:
Yeah, I am actually looking at the third option you mentioned, other than also telling the upstream projects about it.

> Frigate does not send `thumbnail` through the MQTT event, so the file will have to be loaded from somewhere else.

I don't understand this yet. Personally, I've not discriminated between "snapshot image" and "thumbnail" yet, but I am not a user of Frigate. As far as I was concerned, images are arriving now in mqttwarn using the code we added to unblock receiving binary payloads, and that data was what I was looking at. Please educate me if and how this `thumbnail` thing would be any kind of blocker.

sevmonster:
`thumbnail` is a property of the `Event` object in Frigate. It is just a thumbnail of the snapshot. Snapshots are still frames of detection events.

Frigate seems to be excluding `thumbnail` before sending the contents of the `Event` over MQTT, and that's why the current workaround is used with the `snapshot` endpoint—though so far it seems to be working pretty well.

I have no preference on whether the thumbnail or full snapshot (internally and previously known as *clips*) is used. Snapshots are usually very small in resolution anyway.

sevmonster:
As far as I can see the image usually _arrives_ after, though I can't confirm if it is actually being _sent_ after or not. This seems to end up working fine in my environment, as the latency for Ntfy to send the HTTP request seems to take just longer than publishing and saving the new screenshot.

amotl:
> As far as I can see the image usually _arrives_ after.

I've expected the same. However, this detail is important. What does `mosquitt_sub -t 'frigate/#'` reveal?

examples/frigate/frigate.ini
```diff
@@ -0,0 +1,56 @@
...
+
+# Configure `file` plugin to pass through payload 1:1.
+append_newline = False
+decode_utf8    = False
+overwrite      = True
+
+[frigate/cam-testdrive/goat/snapshot]
+targets = store-jpeg:cam-testdrive-goat
+[frigate/cam-testdrive/squirrel/snapshot]
+targets = store-jpeg:cam-testdrive-squirrel
```
amotl, line 56: This section may be condensed, depending on your taste, see [Templated targets » Example 2](https://mqttwarn.readthedocs.io/en/latest/configure/topic.html#example-2).

examples/frigate/frigate.py
amotl: Thanks. I've just added 29e93828b4a6 which addresses your suggestion.

PR #645, opened 2023-04-28
Adjustments to packaging (OCI) and CI configurations

amotl:
### About
Build times of OCI image `mqttwarn-full` increased, probably due to some changes in dependencies. Currently, it needs 10 minutes to build images of that variant for `amd64`, `arm64`, and `armv7` platforms.

### Analysis
I think that the `armv7` builds are mostly responsible for the speed bump, because some of them lack armv7 wheel packages on PyPI.

### Improvement
This patch aims to improve OCI image build times. The outcome is promising.

- Release Docker Standard: 3m 3s
- Release Docker Full: 5m 8s

#### Packaging/OCI
- Skip installing `twilio` on `armv7`, it's too heavy.
- Don't install `build-essential` package until further notice, it is apparently not needed.
- Upgrade `pip` prior to installing other Python packages.
- Add detailed section about image sizes to documentation.

#### CI
- Adjust caching within OCI build configuration, to avoid eventual collisions.

### DIFF
--- a/docs/usage/oci.md
+++ b/docs/usage/oci.md
@@ -205,8 +205,21 @@ If you prefer not to fiddle with those details, but instead want to run a full
 image including dependencies for all modules, we have you covered. Alongside
 the standard image, there is also `ghcr.io/jpmens/mqttwarn-full:latest`.
 
-The `standard` image weighs in with about 130 MB, the `full` image has 230 MB.
+## Image sizes
+
+We determined the **compressed** image sizes using [dockersize]::
+
+    $ dockersize ghcr.io/jpmens/mqttwarn-standard:latest
+    linux/amd64   172.89M
+    linux/arm64   167.12M
+    linux/arm/v7  143.39M
+
+    $ dockersize ghcr.io/jpmens/mqttwarn-full:latest
+    linux/amd64   205.96M
+    linux/arm64   186.81M
+    linux/arm/v7  160.47M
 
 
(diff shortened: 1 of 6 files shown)


docs/usage/oci.md
```diff
@@ -205,8 +205,21 @@ If you prefer not to fiddle with those details, but instead want to run a full
...
+    linux/arm64   167.12M
+    linux/arm/v7  143.39M
+
+    $ dockersize ghcr.io/jpmens/mqttwarn-full:latest
+    linux/amd64   205.96M
+    linux/arm64   186.81M
+    linux/arm/v7  160.47M
+
+
+[dockersize]: https://gist.github.com/MichaelSimons/fb588539dcefd9b5fdf45ba04c302db6?permalink_comment_id=4243739#gistcomment-4243739
```
amotl, line 223: I wonder if that difference of ~30 MB really satisfies shipping two different variants. Given, the delta is larger on the uncompressed image sizes (490 MB vs. 615 MB), but the **compressed size** is what mostly counts, as this is going over the wire, right?

We've only recently tripped into this when spending quite an amount of time aiming to shrink a few Docker images, only looking at the output of, for example, `docker inspect -f "{{ .Size }}" ghcr.io/jpmens/mqttwarn-standard`, and finally discovered that the total compressed image sizes were not significantly smaller than before our optimizations.

jpmens (approved):

amotl:
Runtimes are way better than before now.

image

PR #663, opened 2023-05-07
Combine GitHub Pages configuration with RTD-hosted documentation

amotl:
### About
An attempt to let http://www.mqttwarn.net/ redirect to https://mqttwarn.readthedocs.io/.

### Details
The _GitHub Pages_ settings at [^1] have been adjusted to build from the **repository root** instead of the `docs` subdirectory, which now contains a [Sphinx](https://www.sphinx-doc.org/) source tree instead of Jekyll, as before. As such, the `CNAME` control file has been moved to the repository root folder.

### References
- https://stackoverflow.com/questions/30167113/redirect-github-pages-to-custom-domain
- https://docs.github.com/en/pages/configuring-a-custom-domain-for-your-github-pages-site/managing-a-custom-domain-for-your-github-pages-site

[^1]: https://github.com/mqtt-tools/mqttwarn/settings/pages

PR #664, opened 2023-05-07
Documentation: Move `CNAME` file to `docs-static` folder

amotl:
Following up on GH-663, this moves the `CNAME` control file from the repository root into the `docs-static` folder, to be deployed by/to GitHub Pages. Maybe this way, the HTML-based redirect can be mitigated?


amotl:
> Maybe this way, the HTML-based redirect can be mitigated?

Apparently, it does not work that way.

PR #669, opened 2023-05-13
[udf] Unlock Lua for user-defined functions

amotl:
Hi again,

similar to what [Webscript](https://webscript.io/) has been doing for HTTP, and @progrium explored on behalf of [Airscript](https://github.com/airscript/airscript-engine) the other day, this patch explores the feasibility of "using Lua for MQTT scripting".

Thanks to the excellent [`lupa`](https://github.com/scoder/lupa) package by @scoder 💯, the patch is ridiculously small again, similar to the other one GH-667, which brings in JavaScript for the same reasons. Both patches are making mqttwarn a host for other languages, effectively bringing together the MQTT protocol, mqttwarn's transformation engine, and now, more languages than just Python, into a small-scale framework it is so much fun to fiddle with.

I think in the current era of [serverless computing] and [FaaS] systems, often operated by large cloud service vendors, it is so important to have open source alternatives around, which, more often than not, can do things more efficiently and conveniently in terms of development efforts, and also don't have several layers of indirection between caller and callee, omitting all the dynamic scaling and throttling aspects usually present in large application container / language host infrastructures, which I found to have a _significant_ cost, both mentally and operationally.

Maybe it works well when it's deployed well, but well, long story short, I prefer [personal computing](https://en.wikipedia.org/wiki/Personal_computing) over needing to wrap my head around how to invoke services/programs/functions on 3rd-party infrastructure, which is mostly not completely transparent, or impossible to run on a personal computer at all.

My motivation is mainly but not exclusively driven by my bad experiences in the past when I have been sucked into analyzing subtle errors on shared cloud infrastructure. It's ridiculous, and I am so happy to have learned how to operate my own machines for hosting and publishing digital goods. Enjoy reading! [^1]

- https://github.com/confluentinc/librdkafka/issues/3109

The documentation of the new feature reflects my experiences, and its preview is available here:

- https://mqttwarn--669.org.readthedocs.build/en/669/configure/transformation.html#using-other-languages
- https://mqttwarn--669.org.readthedocs.build/en/669/examples/owntracks-ntfy/readme-variants.html#lua

With kind regards,
Andreas.

[FaaS]: https://en.wikipedia.org/wiki/Cloud_computing#Serverless_computing_or_Function-as-a-Service_(FaaS)
[serverless computing]: https://en.wikipedia.org/wiki/Serverless_computing

[^1]: And no offense to the Azure and Kafka authors, engineers and users, I know I am comparing apples with oranges here, you are doing an excellent job, and nobody would want to professionally run Kafka on personal computers anyway ;]. It's all just my personal idea of a small-scale FaaS framework with an adapter to a message bus system - here, MQTT.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,6 +6,7 @@ mqttwarn changelog
 in progress
 ===========
 - [udf] Unlock JavaScript for user-defined functions. Thanks, @extremeheat.
+- [udf] Unlock Lua for user-defined functions. Thanks, @scoder.
 
 - CI: Validated on Python 3.14
 - OCI: Updated to Python 3.14
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN --mount=type=cache,id=pip,target=/root/.cache/pip \
     true \
     && pip install --upgrade pip \
     && pip install --prefer-binary versioningit wheel \
-    && pip install --use-pep517 --prefer-binary '/src[javascript]'
+    && pip install --use-pep517 --prefer-binary '/src[javascript,lua]'
 
 # Uninstall build prerequisites again.
 RUN apt-get --yes remove --purge git && apt-get --yes autoremove
--- a/docs/configure/transformation.md
+++ b/docs/configure/transformation.md
@@ -447,6 +447,19 @@ export its main entry point symbol, configure mqttwarn to use `functions = myclo
 and adjust its settings to use your MQTT broker endpoint at the beginning of the data
 pipeline, invoke mqttwarn, and turn off Kafka. It works!
 
+On the next day, after investigating if you need to migrate any other system components,
+you realize that there is an Nginx instance, which receives a certain share of telemetry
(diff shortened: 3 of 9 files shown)


amotl:
Hi Stefan. Thanks for the heads up. I had the impression that I was using Lupa 2.0 already. It works very well, thank you so much!

```
$ pip install lupa
Collecting lupa
  Using cached lupa-2.0.tar.gz (6.3 MB)
```

```
$ python -c 'import lupa; print(lupa.__version__)'
2.0
```

scoder:
> I had the impression that I was using Lupa 2.0 already.

Ah, right, sorry. When you maintain multiple packages, it's easy to forget which is released at which version (or not yet).

The simple "import lupa" obviously still works and gives you the latest available Lua version. That might not be the best choice for an embedded scripting engine, since a new lupa release might also change that version. You might want to keep the version import explicit in your usage code, and change it at your own pace.

> It works very well, thank you so much!

Happy to read that.

PR #685, opened 2023-10-14
Add support for Python 3.12

amotl:
## About

Python 3.12 has been released recently. This patch attempts to add corresponding support.

## Details

- For loading code at runtime, use `importlib` instead of `imp`.
  The `imp` module was deprecated and got removed from Python 3.12.
- Use `importlib.resources` instead of deprecated `pkg_resources`.
- Remove dependency on `future` package, for `http` and `pushsafer` service plugins.
- Upgrade to pyfakefs<6.
- Skip installing and testing twilio on Python 3.12.

## Trivia

Start omitting intermediate Python versions on CI, to save resources.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,9 +18,10 @@ jobs:
   tests:
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: ["ubuntu-20.04"]
-        python-version: ["3.6", "3.12"]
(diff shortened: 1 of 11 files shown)


amotl:
[aiohttp](https://pypi.org/project/aiohttp/) needs to be made compatible with Python 3.12 first. It is pulled in by the [twilio](https://pypi.org/project/twilio/) package.
```
Failed to build aiohttp
```

PR #691, opened 2024-02-06
Patch for a shorthand definition of quoted fields in http_urllib

otdftr:
When working with a greater number of transformed fields (e. g. through topic's alldata), the params definition for the target has a lot of redundancy, especially when the field names are crafted to be the parameter names of the query parameters:

This patch provides a shorthand for a definition like 
`[ #method, #url, { 'param1' : '@param1', ..., 'paramN': '@paramN' }, ...`
by allowing lists for the 3rd parameter of the target definition:
`[ #method, #url, [ '?param1', ..., '?paramN' ], ...`

If a list is provided

- the given names will be used as both keys do item.data{} and query parameter names
- they will be added as quoted fields, regardless of whether the @-prefix is provided: `[ 'param1', ..., 'paramN' ]` will be interpreted as `[ '@param1', ..., '@paramN' ]`
- with the prefix `?`, fields can be declared optional: they will not be included in the query if the data is invalid or missing from item.data{}; `[ 'param1', '?param2', ... ]`

### DIFF
--- a/mqttwarn/services/http_urllib.py
+++ b/mqttwarn/services/http_urllib.py
@@ -24,8 +24,8 @@ except ImportError:
 
 def plugin(srv, item):
     """ addrs: (method, url, dict(params), list(username, password), json) 
-        or (shorthand)
-        addrs: (method, url, list(params), list(username, password), json)  """
+        or (shorthand for quoted fields)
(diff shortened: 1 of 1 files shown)


otdftr:
this one hat two bugs, creating new PR

PR #697, opened 2024-07-07
Introduce a warning timeout for MQTT topics

gbeine:
The feature allows to define a parameter 'timeout' for each topic. The timeout is specified in seconds.

Instead of notifying targets when a message arrives for a topic, mqttwarn will trigger the targets if no message is received on the topic for the specified period of time.

That's the first version of this feature, so any feedback is welcome.

### DIFF
--- a/mqttwarn/context.py
+++ b/mqttwarn/context.py
@@ -57,6 +57,12 @@ class RuntimeContext:
             timeout = int(self.config.get(section, "timeout"))
         return timeout
 
+    def get_notify_only_on_timeout(self, section: str) -> int:
+        notify_only_on_timeout = False
+        if self.config.has_option(section, "notify_only_on_timeout"):
+            notify_only_on_timeout = bool(self.config.get(section, "notify_only_on_timeout"))
+        return notify_only_on_timeout
+
     def get_config(self, section: str, name: str) -> t.Any:
         value = None
         if self.config.has_option(section, name):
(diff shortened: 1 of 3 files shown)


amotl:
Hi Gerrit,

apologies for the late reply. That's a nice liveness sensor you are contributing here. I didn't have a look at the details yet, but if @jpmens or @sumnerboy12 don't have any objections about adding that as a new kind of feature, let's try to bring it into mainline.

With kind regards,
Andreas.

gbeine:
@amotl Thank you - and no problem, we're all busy folks ;-)

I'd welcome any feedback as I'm not sure if I've done the implementation in the most effective way.
All I can say is, that it runs on some Raspberry Pi 2 and 4 machines with about 70 to 150 topics stable for months.

gbeine:
Is there a chance to merge this PR?

jpmens:
I think @amotl wanted to comment on some of the design and implementation, but he's not yet had an opportunity to do so.

gbeine:
Thanks for the fast answer @jpmens!
It would be great to see how the feature can be improved.
Meanwhile I have to maintain an own docker image until the feature is merged ;-)

amotl:
Hi again. Thanks for the reminder. I will try to come back to this soon. 🙏

amotl:
> While your implementation looks pretty solid, I may have a few suggestions about design and implementation details as we go.

I think we will be happy to merge as-is, if nobody else has any objections. Maybe add software tests and a changelog item, when possible?

amotl:
Hi again. Your patch has been merged, thank you again and sorry for the delay. When possible, a bit of documentation can be useful for other users. 🙏

amotl:
Hi again. Your improvement has been released with v0.36.0.

PR #703, opened 2026-01-28
Chore: Fix software tests after a hiatus

amotl:
Just maintenance.


amotl:
No idea where this was coming from. Apparently, I was on a super old branch of mqttwarn. Apologies for the noise.

Issue #706, opened 2026-05-08
Desktop-Notifier No Longer Supports Sync Calls


ignorantbliss:
It looks like Desktop-Notifier no longer supports synchronous calls - the **call_sync** function isn't available anymore.

amotl:
Hi @ignorantbliss. Thank you for reporting this. Can you submit a pull request to address the problem?

amotl:
Hi again. I've opened a PR with your patch.

- GH-711

amotl:
Hi @ignorantbliss. We've used your original patch, satisfied software tests, and staged it separately.

- GH-712

amotl:
We released mqttwarn v0.36.1 which includes your improvement, thank you again. Can you verify Desktop-Notifier works again?

amotl:
Gentle ping. ;]

PR #711, opened 2026-07-12
Modernize Desktop-Notifier

amotl:
## About
- Desktop-Notifier no longer supports sync calls.
- Permit working with non-string data (i.e. dictionaries).

## References
- GH-706


amotl:
Deferred to GH-712, because I can't push to the remote branch.

PR #716, opened 2026-07-20
Clarify filter suppression semantics

### LINKED ISSUE #637: Documentation vs. code: Clarify and/or fix logic of "filter" function
Dear JP and Ben,

this is an important issue. Maybe you can find a few minutes to look into it.

With kind regards,
Andreas.

### Introduction

I am bringing this up, because @sevmonster shared their experiences when using mqttwarn's `filter` feature, and because, after comparing it with the documentation, I discovered an anomaly between mqttwarn's documentation and implementation.

> In its current incarnation, should it not be stopping all messages since it always returns `False`? Why are messages getting through? Am I misunderstanding the filter function usage?
>
> I did individually test the logic in the conditional with dummy data so the rest of it should be sound and do what I want, with the understanding that `True` is a passing filter and `False` fails.
>
> -- https://github.com/jpmens/mqttwarn/issues/632#issuecomment-1505310506

### Documentation
While working on GH-635, I am now at the spot where the documentation states:

> A function called from the `filter` property in a topic section needs to return `False` to stop the outbound notification.
>
> -- [Custom functions » Filter functions](https://github.com/jpmens/mqttwarn/blob/0.33.0/HANDBOOK.md#filter-functions)

### Observations
However, while working on GH-632 and GH-635, we discovered that the opposite might be the case. Both code examples actually demonstrate that you need to return `True` to stop the outbound notification.

- [Low battery notifications](https://github.com/jpmens/mqttwarn/blob/0.33.0/HANDBOOK.md#low-battery-notifications)
- [Frigate to Ntfy filter](https://github.com/jpmens/mqttwarn/blob/8434441863d471555c11c6f5ba0b2c0359b342d7/examples/frigate/frigate.py#L99-L112)

Sanjays2402:
Closes #637

Clarifies that filter callbacks are suppression predicates: `True` skips a notification and `False` continues delivery. The user guide now includes a truth table, and the API docstring has regression coverage for the same contract.

Validation: focused contract regression (fails without the docstring fix, passes with it), Ruff, Black, Python compilation, and both diff gates passed.

### DIFF
--- a/docs/configure/transformation.md
+++ b/docs/configure/transformation.md
@@ -487,6 +487,12 @@ process the message.
 The filter function should return `True` if the message should be suppressed, 
 or `False` if the message should be processed. This is a basic but working
 example for such a filter function.
+
+| Callback result | Outcome |
+| --- | --- |
+| `True` | Message is filtered out; notification is skipped |
+| `False` | Message proceeds to the configured targets |
+
 ```python
 from mqttwarn.model import Service
 
--- a/mqttwarn/context.py
+++ b/mqttwarn/context.py
@@ -211,12 +211,15 @@ class FunctionInvoker:
     def filter(self, name: str, topic: str, payload: t.AnyStr, section: t.Optional[str] = None) -> bool:  # noqa:A003
         """
         Invoke function "name" loaded from the "functions" Python module.
-        Return that function's True/False.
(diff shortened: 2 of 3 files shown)


amotl (approved):
Thank you very much. 🌻

